PR #4, opened 2023-03-31
Split Styles into three layers

MinusGix:
I think this is maybe ready enough, but I'd be fine with comments on things to change.
- Splits `Style` into three different 'layers', the doc comment at the top of style.rs explains it a bit.
  - These are generated with a macro to avoid repetitive typing and avoid mistakes (like accidentally skipping a field in a function that uses them all)
  - The builder functions are generated as well. I could easily make the macro let you customize their builder function names if we want shorter ones for any of them?
- Has so the actual style for rendering is the `ReifiedStyle` (which you then convert into a Taffy style) formed from the declared `Style`, any override styles (like for hover), and the underlying view style (which, for all views is just the default style like before).  
- Removed the `border`, `padding`, and `margin` fields. There is now only the top/bottom/left/right fields that can be set with functions on `Style` and friends.
- Reworks the examples to use the new `Style`. While you could write it out in a struct like before, having everything be an `Option<T>` makes it more verbose. Also, the builder style is basically the same for all 3 style kinds.

----

- There's still a todo about what to do for `is_hidden`. We don't hit this as a problem right now since no views have their own view-style  and there's no override styles, but it could become an issue..

### DIFF
--- a/examples/counter/src/main.rs
+++ b/examples/counter/src/main.rs
@@ -4,7 +4,7 @@ use floem::{
     peniko::Color,
     reactive::{create_signal, SignalGet, SignalUpdate},
     stack::stack,
-    style::{AlignItems, Dimension, FlexDirection, Position, Style},
+    style::{AlignItems, Dimension, FlexDirection, Style},
     view::View,
     views::Decorators,
     views::{click, label, rich_text, scroll, VirtualListDirection, VirtualListItemSize},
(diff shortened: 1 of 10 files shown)


dzhou121:
I like the macro to generate builder functions. One little request is separating things like width and width_percentage so that you can just put raw value in without the whole Dimension::Points()

dzhou121:
Another thought is can we probably use the same struct for Style and ReifiedStyle, The None in Style being Unset in ReifiedStyle.

PR #9, opened 2023-04-18
Add hover support

### LINKED ISSUE #5: Hover Support
It would be good to have an easy way to have things happen on hover.  
Implementation Suggestions:
[`AppState`](https://github.com/lapce/floem/blob/main/src/context.rs#L89) can have a `hover: Vec<Id>` which is just a list of all the currently hovered views. 
[`ViewState`](https://github.com/lapce/floem/blob/main/src/context.rs#L45) Can get a `hover_style: Option<Style>` (or just `Style::default()`) which is then applied in `fill_reified_style`, if the element is hovered.   
  
Then needs some code to detect when the mouse enters  
```rust
let rect = cx.get_size(self.id).unwrap_or_default().to_rect();
if rect.contains(event.pos) {
```
and set hover, and clear reified style cache  
And some code to detect when the mouse leaves, removing it from the hover list if it is in there and clearing reified style cache.  
Unsure where that logic should go.

presiyan-ivanov:
Closes #5

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -66,14 +66,15 @@ impl ViewState {
         view_style: &ReifiedStyle,
     ) -> &ReifiedStyle {
         let base_style = self.style.clone().reify(view_style);
-        self.reified_style = match interact_state {
-            InteractionState::Hover => Some(
+        self.reified_style = if interact_state.is_hovered {
+            Some(
                 self.hover_style
                     .clone()
                     .unwrap_or_default()
                     .reify(&base_style),
-            ),
-            InteractionState::Default => Some(base_style),
+            )
+        } else {
+            Some(base_style)
         };
 
         self.reified_style.as_ref().unwrap()
@@ -130,10 +131,10 @@ impl AppState {
 
     pub fn get_interact_state(&self, id: &Id) -> InteractionState {
         if self.hovered.contains(id) {
-            return InteractionState::Hover;
+            return InteractionState { is_hovered: true };
         }
 
-        InteractionState::Default
+        InteractionState::default()
     }
 
     pub fn set_root_size(&mut self, size: Size) {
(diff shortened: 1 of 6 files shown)


src/context.rs
```diff
@@ -227,12 +260,22 @@ impl<'a> EventCx<'a> {
...
             }
         }
         false
     }
 }
 
+pub enum InteractionState {
+    Default,
+    Hover,
+    // TODO: focus, active, disabled
```
dzhou121, line 276: What if the ```View``` is both focused and hovered?

presiyan-ivanov: I went with this approach because I think its simpler and cleaner for computing the styles - focus simply will have a higher priority than hover. 
I think if the goal is to support combining interaction states for the styles(e.g. style for hover+focus) then the flag approach is the way  - but it also adds more complexity. For example `is_disabled` needs to be checked before applying other flags, because it doesn't make sense to have a disabled view that is active/focused etc.

dzhou121: I think the flags are useful for other stuff, e.g., ```is_disabled``` can be used by ```event```. 

Also, focus and active are already taken care of in ```AppState```. 

Back to the style, I'm not sure if we should combine all of the states or only the higher priority state. Do you know how web does that?

presiyan-ivanov: In web the styles are combined and the higher priority selectors override the lower priority, but you can also combine selectors. 
Suppose an input is hovered **and** focused -  `:hover:focus` applies its styles on top of  `:focus`  which overrides `:hover` which overrides the base styles.

My concern is if we do it this way, how will the user API work for combining the states(e.g. hover + focus) when styling views? In web its quite easy to combine them (since they are just strings - "input:hover:focus" etc.). Maybe the interactive state can be exposed, so it would be possible to do like so :
 ```rust
 stack()
style(......),
hover_style(.....),
focus_style(......)
chained_style(cx, [
    (InteractionState.Hovered, InteractionState.Focused, || Style::default().background(Color::YELLOW)), 
    (InteractionState.Disabled, InteractionState.Hovered, || Style::default().background(Color::BLACK))
   // the latter one overrides previous, if both match 
]);
 ```

dzhou121: It's quite complex :) I guess we'll worry about the combined selectors later.

If you don't mind changing the hover to be a flag, and we get this RP merged first.

PR #11, opened 2023-04-19
Fix close window

SunDoge:
Now we can close the window by clicking the close button. However, I'm not sure whether the `WindowClosed` event should be placed in the `fn destroy` or the `fn request_close`.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -66,7 +66,10 @@ impl AppContext {
     }
 
     pub fn update_hover_style(id: Id, style: Style) {
-        UPDATE_MESSAGES.with(|msgs| msgs.borrow_mut().push(UpdateMessage::HoverStyle { id, style }));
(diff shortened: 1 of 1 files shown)


dzhou121:
One thing I need to quickly point out is that Mac doesn't quit the app when the windows are closed.

SunDoge:
Unfortunately, I don't have a MacOS device for testing. On Linux, `Application::global().quit()` can make the app exit after the window is closed. This code is from xilem [https://github.com/linebender/xilem/blob/main/src/app_main.rs#L169](https://github.com/linebender/xilem/blob/main/src/app_main.rs#L169). Is xilem works properly on MacOS?

dzhou121:
I guess I can merge this first and refine it later, if you could resolve the conflict.

SunDoge:
Done.

PR #14, opened 2023-04-21
fix: allow mousemove event processing for disabled view

bbrabbitt:
The current counter example contains a bug that after clicking reset and click on other buttons will cause reset button to react.

Since in current code when a view is disabled it will not process `MouseMove` event, which in turn prevents itself from removing from the hovered list. The `MouseMove` event will only be processed when the button is not disabled. In counter example this is when either Increment or Decrement is clicked.

The fix solves the problem by adding MouseMove to `allow_disabled` list.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -49,10 +49,9 @@ impl Event {
         match self {
             Event::MouseDown(_)
             | Event::MouseUp(_)
-            | Event::MouseMove(_)
             | Event::MouseWheel(_)
             | Event::KeyDown(_) => false,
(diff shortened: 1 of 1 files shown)


dzhou121:
Are you on the latest master. I feel like I’ve fixed this issue by removing the view from hovered list when it’s disabled.

bbrabbitt:
> Are you on the latest master. I feel like I’ve fixed this issue by removing the view from hovered list when it’s disabled.

Oh right I'm not on the latest. The latest master doesn't have the problem. I'll close the PR.

PR #19, opened 2023-05-02
Feat/add text input

presiyan-ivanov:
Adds support for single line text input. There are lots of things missing at the moment(text selection, copy/paste/undo/redo, word movement, etc.) but it is still somewhat usable. We can create tickets for the other things, there are a lot of small things that can be a good first issue for someone


src/views/text_input.rs
```diff
@@ -0,0 +1,491 @@
...
+    id: Id,
+    buffer: RwSignal<String>,
+    // Where are we in the main buffer
+    cursor_glyph_idx: usize,
+    // This can be retrieved from the glyph, but we store it for efficiency
+    cursor_x: f64,
+    text_buf: Option<TextLayout>,
+    text_node: Option<Node>,
+    // Shown when the width exceeds node width for single line input
+    clipped_text: Option<String>,
```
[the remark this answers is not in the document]
presiyan-ivanov: Do you mean  using the `cosmic_text::TextLayout::set_scroll` API or just break the text to new lines every time it exceeds widthand the user has to scroll? :D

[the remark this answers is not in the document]
dzhou121: I meant the ```scroll``` view in Floem.

src/style.rs
```diff
@@ -238,7 +238,7 @@ define_styles!(
     cursor nocb: Option<CursorStyle> = None,
     color nocb: Option<Color> = None,
     background nocb: Option<Color> = None,
-    font_size nocb: Option<f32> = None,
+    font_size nocb: f32 = 14.0,
```
dzhou121, line 241: With ```Option<f32>```, I could set a font size at the parent, and the children will have the same font size. With this, the default style will override the parent's font size.

presiyan-ivanov: Oof, I see. I will change it back then.

src/view.rs
```diff
@@ -60,9 +60,7 @@ pub trait View {
         cx.app_state.compute_style(self.id(), view_style);
         let style = cx.app_state.get_computed_style(self.id()).clone();
 
-        if style.font_size.is_some() {
-            cx.font_size = style.font_size;
-        }
+        cx.font_size = style.font_size;
```
dzhou121, line 63: It needs to be reverted here as well.

src/view.rs
```diff
@@ -217,9 +227,7 @@ pub trait View {
             if style.color.is_some() {
                 cx.color = style.color;
             }
-            if style.font_size.is_some() {
-                cx.font_size = style.font_size;
-            }
+            cx.font_size = style.font_size;
```
dzhou121, line 230: And here.

PR #36, opened 2023-05-12
Implment View for () for an empty type

jrmoulton:
This is done in order to have an empy type that implements view. This makes it so that a container can have nothing inside of it but can still display properties like border. This is useful for creating shapes.

I'm not sure where the best place to put this code would be so I just implemented it right beneath the View trait. 

An example where this is useful is creating a divider
``` rust
container(|| ()).style(|| {
    Style::BASE
        .width_px(25.0)
        .height_px(0.0)
        .border_color(Color::WHITE)
        .border(1.5)
        .border_radius(3.0)
}),
```

### DIFF
--- a/src/view.rs
+++ b/src/view.rs
@@ -455,34 +455,3 @@ fn paint_border(cx: &mut PaintCx, style: &ComputedStyle, size: Size) {
         }
     }
 }
-
-impl View for () {
-    fn id(&self) -> Id {
-        let cx = crate::AppContext::get_current();
-        cx.new_id()
-    }
-
-    fn child(&mut self, _id: Id) -> Option<&mut dyn View> {
-        None
-    }
-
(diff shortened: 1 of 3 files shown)


dzhou121:
It would be perfect if we could do 
```rust
().style(|| {
    Style::BASE
        .width_px(25.0)
        .height_px(0.0)
        .border_color(Color::WHITE)
        .border(1.5)
        .border_radius(3.0)
```

The main issue is that if we could find a way to have a static id with it. 

Failing that, I'm wondering if a new "empty"(or probably a better name) view would be better, e.g.

```rust
empty().style(|| {
    Style::BASE
        .width_px(25.0)
        .height_px(0.0)
        .border_color(Color::WHITE)
        .border(1.5)
        .border_radius(3.0)
})
```

jrmoulton:
I've gone with the empty() approach but I couldn't come up with a better name

PR #37, opened 2023-05-13
Support drawing circles

jrmoulton:
I wasn't sure the best way to implement this. It would make sense to support ellipses instead of circles (that would work better with both width and height properties) but this seemed to be what is currently supported in vger.

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -228,6 +228,7 @@ define_styles!(
     border_bottom: f32 = 0.0,
     border_radius: f32 = 0.0,
     border_color: Color = Color::BLACK,
+    circle: bool = false,
     padding_left: LengthPercentage = LengthPercentage::ZERO,
     padding_top: LengthPercentage = LengthPercentage::ZERO,
     padding_right: LengthPercentage = LengthPercentage::ZERO,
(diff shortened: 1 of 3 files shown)


dzhou121:
I'm wondering if it's probably cleaner to do the css approach, e.g., it became a circle when the border-radius is greater than 50%.

jrmoulton:
This is what I initially tried to do but kurbo has a cap on the amount of border radius that can be set

dzhou121:
Yeah I tried it and it didn't work. 

We can probably just create a circle "manually" with a ```if border_radis >= 50%```

jrmoulton:
Ahh yeah that would be better. It still has the problem of not taking into account both width and height but that's at least better

jrmoulton:
I'm starting to think that the problem is actually in vger. The rounded rectangles in kurbo are just fine. My initial guess is that the issue is here
https://github.com/lapce/vger-rs/blob/aa84d85c9f7df9da6ef47a2b6d43fe6909023b9b/src/shader.wgsl#LL106C1-L110C2

jrmoulton:
I just updated to the macOS beta and this is no longer an issue. Border radius does actually change the border radius the way it should

PR #38, opened 2023-05-13
Add PointerEnter and PointerLeave EventListeners

jrmoulton:
This adds support for a pointer entering and leaving. This does work but I'm marking this as a draft because this doesn't run inside of the `event.needs_focus()` check but I don't fully understand what that is.


dzhou121:
```event.needs_focus()``` basically is to check if the event is a keyboard event.

jrmoulton:
> event.needs_focus() basically is to check if the event is a keyboard event.

Ok. In that case the way it is should work fine

PR #40, opened 2023-05-15
Add WindowDragArea

jrmoulton:
This is a TitleBar view to be used as a handle to move the application when the native title bar is disabled. 

This works but breaks when the mouse moves faster than the window and the move event doesn't get sent to the view anymore.

### DIFF
--- a/src/app_handle.rs
+++ b/src/app_handle.rs
@@ -686,10 +686,9 @@ impl<V: View> WinHandler for AppHandle<V> {
             MousePosState::None => {}
             MousePosState::Ready => self.prev_mouse_pos = MousePosState::Some(event.pos),
             MousePosState::Some(prev_pos) => {
-                self.handle
-                    .set_position(Point::from(std::convert::Into::<(f64, f64)>::into(
-                        self.handle.get_position() + (event.pos - prev_pos),
-                    )));
+                let position_diff = event.pos - prev_pos;
+                let new_position = self.handle.get_position() + position_diff;
+                self.handle.set_position(new_position);
             }
         }
         self.event(Event::PointerMove(event.clone()));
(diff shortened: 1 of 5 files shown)


jrmoulton:
This is working now

src/app_handle.rs
```diff
@@ -577,10 +594,20 @@ impl<V: View> WinHandler for AppHandle<V> {
...
     }
 
     fn pointer_move(&mut self, event: &glazier::PointerEvent) {
+        match self.prev_mouse_pos {
+            MousePosState::None => {}
+            MousePosState::Ready => self.prev_mouse_pos = MousePosState::Some(event.pos),
+            MousePosState::Some(prev_pos) => {
+                let position_diff = event.pos - prev_pos;
+                let new_position = self.handle.get_position() + position_diff;
+                self.handle.set_position(new_position);
```
dzhou121, line 608: I feel like after ```handle_titlebar(true)```, the OS will take care of the window move.

It is only implemented on Windows in Glazier though.

jrmoulton: It doesn't. I've tested it on windows and it doesn't move. I think it just lets the OS know that it's a titlebar for things like window snapping

jrmoulton: Although window snapping on windows isn't working either so I'm not sure

jrmoulton:
Any thoughts or feelings on the current implementation of this PR?

dzhou121:
It's only the name of the view I'm not sure of.

I'm thinking ```window_drag_area``` or maybe just ```handle_titlebar```?

jrmoulton:
window_drag_area would probably be good.

Also any idea why window snapping in windows doesn't work with the way I have things implemented? (Window snapping being dragging a window the the side of the screen and having it expand to fill a section of the screen).

dzhou121:
It seems the snapping is lost when titlebar is hidden.

dzhou121:
In fact, you can test to see if it works with titlebar on.

jrmoulton:
The window snapping still isn't working so there are issues to work out but this should be good to be merged

PR #48, opened 2023-05-29
remove opacity setting from scroll bars

jrmoulton:
No need for an opacity setting as peniko::Color already support opacity

### DIFF
--- a/src/views/scroll.rs
+++ b/src/views/scroll.rs
@@ -57,7 +57,6 @@ pub struct Scroll<V: View> {
     held: BarHeldState,
     virtual_node: Option<Node>,
     hide_bar: bool,
-    opacity: f64,
     scroll_bar_color: Color,
 }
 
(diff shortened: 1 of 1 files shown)


jrmoulton:
Optionally the default color could be created using `scroll_bar_color: Color::rgb8(0, 0, 0).with_alpha_factor(0.7)` but then it isn't all const fn but that probably doesn't really matter

jrmoulton:
Good as is or should it change?

dzhou121:
Sorry for looking at this so late. The opacity was there for the fade out of the scroll bar which isn't implemented yet.

jrmoulton:
> Sorry for looking at this so late. 
No worries. 

> The opacity was there for the fade out of the scroll bar which isn't implemented yet. 

Right. But scroll does accept a color for the scroll bar and that color should be what stores the opacity settings and not an extra field. Since peniko Color already supports opacity. This way if you pass in a color with higher or lower opacity it will change how the scroll bar looks which makes sense. Is there any reason to not store the opacity in the color?

dzhou121:
Now thinking it again, it makes sense now.

PR #50, opened 2023-06-03
set text_input width

jrmoulton:
This will set the text input width to the width of the text instead of just 10*font size. 

I've also included a change to the pointer move event. It now return false when it handles the changing of the cursor style so that if an event listener is placed on the text_input it will still get processed


jrmoulton:
This currently has some strange interactions with text clipping

jrmoulton:
The last commit I pushed fixes the issue with text clipping but it does that by just requesting a layout if the width changes. Is there a better way to do that?

jrmoulton:
> If the idea is to set width of visible text via the styles, then we need another approach most likely

Ahh I see. Yeah this is what I'd be looking for

jrmoulton:
I had specially made that change because it fixed centering the element but it had issues so I just left it as a draft

PR #54, opened 2023-06-06
Add link

jrmoulton:
This adds a dependency on [open](https://crates.io/crates/open)

I'm not sure that I love this implementation because it doesn't allow the user to configure when the event happens or even which app should be used to open the link but this was the simplest method. 

Any thoughts?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ im = "15.1.0"
 parking_lot = { version = "0.12.1" }
 floem_renderer = { path = "renderer" }
 floem_vger = { path = "vger" }
+open = "4.1.0"
 
 [workspace]
 members = ["renderer", "vger", "examples/*"]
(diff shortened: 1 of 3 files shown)


dzhou121:
I'm wondering if this is something a user can just implement themselves. with ```on_click```, they can easily achieve this.

jrmoulton:
It definitely can be. That's how I was doing it before. In a lot of ways it's better for floem to not provide this because it makes customization hard. I added it though because it does seem like most UI frameworks have a quick and easy way to add a link with some sane defaults. 

If this did get merged it would probably be a good idea to reexpose `open` so that users could build their own just by importing the right functions. 

If we don't merge this then I'd probably at least add some docs that recommend using open and maybe give an example of how to build a custom link.

jrmoulton:
closing for now

Issue #57, opened 2023-06-10
Textinput causes "Access a signal or memo outside a reactive tracking context" warning 


matthewgapp:
Is this expected? I'm new to leptos. To reproduce, just `cargo run` on the widget gallery example. 

```shell
At src/views/text_input.rs:326:14, you access a signal or memo (defined at examples/widget-gallery/src/inputs.rs:14:16) outside a reactive tracking context. This might mean your app is not responding to changes in signal values in the way you expect.

Here’s how to fix it:

1. If this is inside a `view!` macro, make sure you are passing a function, not a value.
  ❌ NO  <p>{x.get() * 2}</p>
  ✅ YES <p>{move || x.get() * 2}</p>

2. If it’s in the body of a component, try wrapping this access in a closure:
  ❌ NO  let y = x.get() * 2
  ✅ YES let y = move || x.get() * 2.

3. If you’re *trying* to access the value without tracking, use `.get_untracked()` or `.with_untracked()` instead.
At src/views/text_input.rs:326:14, you access a signal or memo (defined at examples/widget-gallery/src/inputs.rs:14:16) outside a reactive tracking context. This might mean your app is not responding to changes in signal values in the way you expect.
```

I can reproduce with this minimal code snippet: 

```rust
struct Repro {
    id: Id,
    text: RwSignal<String>,
    child: TextInput,
}
fn repro() -> Repro {
    let cx = AppContext::get_current();
    let text = create_rw_signal(cx.scope, "Hello World!".to_string());
    let id = cx.new_id();
    let mut child_cx = cx;
    child_cx.id = id;
    AppContext::save();
    AppContext::set_current(child_cx);
    let child = text_input(text);
    AppContext::restore();
    Repro { id, child, text }
}

impl View for Repro {
    fn id(&self) -> Id {
        self.id
    }

    fn child(&mut self, id: Id) -> Option<&mut dyn View> {
        if self.child.id() == id {
            Some(&mut self.child)
        } else {
            None
        }
    }

    fn children(&mut self) -> Vec<&mut dyn View> {
        vec![&mut self.child]
    }

    fn debug_name(&self) -> std::borrow::Cow<'static, str> {
        "Repro".into()
    }

    fn update(&mut self, _cx: &mut UpdateCx, _state: Box<dyn std::any::Any>) -> ChangeFlags {
        ChangeFlags::empty()
    }

    fn layout(&mut self, cx: &mut LayoutCx) -> Node {
[...]

PR #58, opened 2023-06-10
fix warnings by converting `with` to `with_untracked`

matthewgapp:
This PR fixes the warning: `At src/views/text_input.rs:326:14, you access a signal or memo (defined at examples/widget-gallery/src/inputs.rs:14:16) outside a reactive tracking context. This might mean your app is not responding to changes in signal values in the way you expect.`

which is reproducible when running the widget gallery example. 

I'm new to Leptos but seems like it's safe to assume that the locations I've replaced are indeed outside of a tracking context and therefore should use the appropriate method. 

Resolves https://github.com/lapce/floem/issues/57


dzhou121:
Thanks for finding the issue and fixing it!

PR #59, opened 2023-06-10
Consolidate constructor stacked context pattern into helper method

matthewgapp:
As I was building my own `View` components, I noticed that I frequently made mistakes when configuring the app context when constructing a child. This PR wraps the id creation and the sandwiching of save/restoring of app context around child the constructor into a single helper method. 

Very open to better naming or a different pattern. But I do think that the pushing/popping of the app context should be done in a single place to remove possibility that `View` authors mess up the ordering.

### DIFF
--- a/src/app_handle.rs
+++ b/src/app_handle.rs
@@ -87,6 +87,13 @@ impl AppContext {
         })
     }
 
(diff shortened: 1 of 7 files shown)


dzhou121:
This is a good one. Thanks!

PR #60, opened 2023-06-11
Improve documentation

matthewgapp:
I think documentation is super important to ramping newcomers. This PR is the first of a couple that I plan on making that aims to improve documentation. It's also littered with spelling fixes and a couple TODOs that I'll clean up once I get answers to the contained questions.  

I've commented where I have questions. Apologies if I don't get everything right. Just started with the project and I'm learning along the way! 

Screenshot of lib.rs doc comment. I'd love to host Floem's rust docs on github pages in a separate PR. 

image

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -14,10 +14,11 @@ pub enum AppEvent {
 }
 
 /// Floem top level application
+/// This is the entry point of the application.
 pub struct Application {
     application: glazier::Application,
     scope: Scope,
-    event_listner: Option<Box<AppEventCallback>>,
+    event_listener: Option<Box<AppEventCallback>>,
 }
 
 impl Default for Application {
@@ -30,7 +31,7 @@ impl glazier::AppHandler for Application {
     fn command(&mut self, _id: u32) {}
 
     fn will_terminate(&mut self) {
-        if let Some(action) = self.event_listner.as_ref() {
+        if let Some(action) = self.event_listener.as_ref() {
             action(&AppEvent::WillTerminate);
         }
     }
@@ -43,7 +44,7 @@ impl Application {
         Self {
             scope,
             application: glazier::Application::new().unwrap(),
-            event_listner: None,
+            event_listener: None,
         }
     }
 
@@ -52,7 +53,7 @@ impl Application {
     }
 
     pub fn on_event(mut self, action: impl Fn(&AppEvent) + 'static) -> Self {
-        self.event_listner = Some(Box::new(action));
+        self.event_listener = Some(Box::new(action));
         self
     }
 
--- a/src/app_handle.rs
+++ b/src/app_handle.rs
@@ -196,7 +196,7 @@ pub enum UpdateMessage {
     },
 }
 
-/// This is the top-level handle that is passed into the backend interface (e.g. `glazier`) to consume and react to window events.
+/// The top-level handle that is passed into the backend interface (e.g. `glazier`) to interact to window events.
 /// Meant only for use with the root view of the application.
 pub struct AppHandle<V: View> {
     scope: Scope,
--- a/src/app_handle.rs
+++ b/src/app_handle.rs
@@ -36,7 +36,6 @@ type DeferredUpdateMessages = HashMap<Id, Vec<(Id, Box<dyn Any>)>>;
 
 // Primarily used to mint and assign a unique ID to each view.
 // It also contains a constant scope which is used to create signals.
-// QUESTION: would it be better to call this `TreeContext`? Since it is really only about minting Ids given the current position in the tree.
 #[derive(Copy, Clone)]
 pub struct AppContext {
     /// used to create new signals
@@ -193,6 +192,10 @@ pub enum UpdateMessage {
 
 /// The top-level handle that is passed into the backend interface (e.g. `glazier`) to interact to window events.
 /// Meant only for use with the root view of the application.
+/// Owns the `AppState` and is responsible for
+/// - processing all requests to update the AppState from the reactive system
+/// - processing all requests to update the animation state from the reactive system
+/// - requesting a new animation frame from the backend
 pub struct AppHandle<V: View> {
     scope: Scope,
     view: V,
@@ -831,7 +834,7 @@ impl<V: View> WinHandler for AppHandle<V> {
     }
 
     fn size(&mut self, size: glazier::kurbo::Size) {
-        self.app_state.update_scr_size_breakpt(size);
+        self.app_state.update_screen_size_bp(size);
         self.event(Event::WindowResized(size));
         let scale = self.handle.get_scale().unwrap_or_default();
         let scale = Scale::new(
--- a/src/app_handle.rs
+++ b/src/app_handle.rs
@@ -16,35 +16,30 @@ use crate::{
     },
     event::{Event, EventListener},
     ext_event::EXT_EVENT_HANDLER,
-    id::{Id, IDPATHS},
+    id::{Id, ID_PATHS},
     responsive::ScreenSize,
     style::{CursorStyle, Style},
     view::{ChangeFlags, View},
 };
 
 thread_local! {
+    /// Stores a queue of update messages for each view. This is a list of build in messages, including a built-in State message
+    /// that you can use to send a state update to a view.
     pub(crate) static UPDATE_MESSAGES: std::cell::RefCell<HashMap<Id, Vec<UpdateMessage>>> = Default::default();
     pub(crate) static ANIM_UPDATE_MESSAGES: std::cell::RefCell<Vec<AnimUpdateMsg>> = Default::default();
+    ///
     pub(crate) static DEFERRED_UPDATE_MESSAGES: std::cell::RefCell<DeferredUpdateMessages> = Default::default();
 }
 
 pub type FileDialogs = HashMap<FileDialogToken, Box<dyn Fn(Option<FileInfo>)>>;
 type DeferredUpdateMessages = HashMap<Id, Vec<(Id, Box<dyn Any>)>>;
 
-/// The top-level handle that is passed into the backend interface (e.g. `glazier`) to interact to window events.
-/// Meant only for use with the root view of the application.
-pub struct AppHandle<V: View> {
-    scope: Scope,
-    view: V,
-    handle: glazier::WindowHandle,
-    app_state: AppState,
-    paint_state: PaintState,
-
-    file_dialogs: FileDialogs,
-}
-
+// Primarily used to mint and assign a unique ID to each view.
+// It also contains a constant scope which is used to create signals.
+// TODO: would it be better to call this `TreeContext`? Since it is really only about minting Ids given the current position in the tree.
 #[derive(Copy, Clone)]
 pub struct AppContext {
+    /// used to create new signals
     pub scope: Scope,
     pub id: Id,
 }
@@ -161,7 +156,7 @@ pub enum UpdateMessage {
         selector: StyleSelector,
         style: Style,
     },
-    KeyboardNavigatable {
+    KeyboardNavigable {
         id: Id,
     },
     Draggable {
@@ -200,12 +195,10 @@ pub enum UpdateMessage {
 /// Meant only for use with the root view of the application.
 pub struct AppHandle<V: View> {
     scope: Scope,
-    // the root view of the application
     view: V,
     handle: glazier::WindowHandle,
     app_state: AppState,
     paint_state: PaintState,
-    prev_mouse_pos: MousePosState,
 
     file_dialogs: FileDialogs,
 }
@@ -416,7 +409,7 @@ impl<V: View> AppHandle<V> {
             app_state: &mut self.app_state,
         };
         for (id, state) in msgs {
-            let id_path = IDPATHS.with(|paths| paths.borrow().get(&id).cloned());
+            let id_path = ID_PATHS.with(|paths| paths.borrow().get(&id).cloned());
             if let Some(id_path) = id_path {
                 flags |= self.view.update_main(&mut cx, &id_path.0, state);
             }
@@ -490,7 +483,7 @@ impl<V: View> AppHandle<V> {
                         cx.app_state.request_layout(id);
                     }
                     UpdateMessage::State { id, state } => {
-                        let id_path = IDPATHS.with(|paths| paths.borrow().get(&id).cloned());
+                        let id_path = ID_PATHS.with(|paths| paths.borrow().get(&id).cloned());
                         if let Some(id_path) = id_path {
                             flags |= self.view.update_main(&mut cx, &id_path.0, state);
                         }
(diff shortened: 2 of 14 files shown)


src/app.rs
```diff
@@ -43,7 +44,7 @@ impl Application {
         Self {
             scope,
             application: glazier::Application::new().unwrap(),
-            event_listner: None,
+            event_listener: None,
```
matthewgapp, line 47: spelling

src/app.rs
matthewgapp, line 1: changes in this file are just spelling

src/app_handle.rs
```diff
@@ -16,33 +16,30 @@ use crate::{
     },
     event::{Event, EventListener},
     ext_event::EXT_EVENT_HANDLER,
-    id::{Id, IDPATHS},
+    id::{Id, ID_PATHS},
```
matthewgapp, line 19: spelling

src/app_handle.rs
```diff
@@ -16,33 +16,30 @@ use crate::{
...
-    handle: glazier::WindowHandle,
-    app_state: AppState,
-    paint_state: PaintState,
-
-    file_dialogs: FileDialogs,
-}
-
+// Primarily used to mint and assign a unique ID to each view.
+// It also contains a constant scope which is used to create signals.
+// TODO: would it be better to call this `TreeContext`? Since it is really only about minting Ids given the current position in the tree.
```
matthewgapp, line 39: I'd like to call this something like `TreeContext` since it is really only about minting Ids given the current position in the tree.

src/app_handle.rs
```diff
@@ -16,33 +16,30 @@ use crate::{
...
 #[derive(Copy, Clone)]
 pub struct AppContext {
+    /// used to create new signals
```
matthewgapp, line 42: it might even make sense to create methods on this struct that create signals?

src/app_handle.rs
```diff
@@ -175,6 +172,18 @@ pub enum UpdateMessage {
     },
 }
 
+/// The top-level handle that is passed into the backend interface (e.g. `glazier`) to interact to window events.
+/// Meant only for use with the root view of the application.
+pub struct AppHandle<V: View> {
```
matthewgapp, line 177: moved to be near implementation

src/context.rs
```diff
@@ -604,6 +614,9 @@ impl<'a> EventCx<'a> {
         }
     }
 
+    /// Used to determine if you should send an event to another view. This is basically a check for pointer events to see if the pointer is inside a child view and to make sure the current view isn't hidden or disabled.
+    /// Usually this is used if you want to propagate an event to a child view
+    // TODO: should we make this public so that it can be used in custom components? It's not a public function atm but I see it used in all of the container-like `View`s.
     pub(crate) fn should_send(&mut self, id: Id, event: &Event) -> bool {
```
matthewgapp, line 614: should we make this public so that it can be used in custom components? It's not a public function atm but I see it used in all of the container-like `View`s.

src/context.rs
```diff
@@ -639,7 +652,10 @@ pub struct InteractionState {
     pub(crate) using_keyboard_navigation: bool,
 }
 
+/// Holds current layout state for given position in the tree.
+/// You'll use this in the `View::layout` implementation to call `layout_node` on children and to access any font
 pub struct LayoutCx<'a> {
+    // TODO: seems like custom widgets like scroll, tab, text_input and more need access to this. Should we make this a public method on LayoutCx?
```
matthewgapp, line 652: seems like custom widgets like scroll, tab, text_input and more need access to this. Should we make this a public method on LayoutCx?

src/lib.rs
```diff
@@ -1,3 +1,119 @@
...
+//! }
+//!
+//! // Usage
+//! fn main() {
+//!     floem::launch(parent(child));
+//! }
+//!
+//! ```
+//!
+//! Global state TBD
```
matthewgapp, line 77: is there an Floem-idiomatic way to share global state?

src/lib.rs
```diff
@@ -1,3 +1,109 @@
...
+//!
+//! # Styling
+//! Floem sizing and positioning layout system is based on the flexbox model using Taffy as the layout engine.
+//! Styles are divided into two parts:
```
matthewgapp, line 81: is this still the latest? From what I can tell, the styling implementation has been moving quite a bit recently.

src/lib.rs
```diff
@@ -1,3 +1,109 @@
...
+//! potentially discuss update messages, deferred update messages, animations
+//!
+//! ## Notable invariants and tolerances
+//! - There can be only one root `View`
+//! - Only one view can be active at a time.
+//! - Only one view can be focused at a time.
+//! - Any other important invariants?
+//!
+//! # Terminology
+//! - **Active view**: The view that can receive mouse events even if the mouse is outside its bounds.
```
matthewgapp, line 101: I don't fully understand what the difference between active and focused is.

src/lib.rs
```diff
@@ -1,3 +1,109 @@
+//! # Floem
+//! Floem is cross-platform GUI framework for Rust 🦀. It aims to be extremely performant while providing world-class developer ergonomics.
+//!
+//! ## Views
+//! Floem models the UI using a tree of [Views](view::View)s that are built once. `Views` react to state changes and events.
+//! Views are type-erased, self-contained components that can be composed together to create complex UIs.
```
jrmoulton, line 6: I'm pretty sure that that views are not type erased. Only when using box_container are views cast to dyn view. All the other times they are strongly typed

src/lib.rs
jrmoulton: Leptos has provide_context and use_context

src/context.rs
jrmoulton: +1 for this. I've resorted to adding event listeners for all of the pointer events that just return true to stop the propagation of events

src/lib.rs
matthewgapp: thank you. meant to say they're not generic over their children or state like some other libs. But think we can leave that out.

src/view.rs
```diff
@@ -163,6 +183,7 @@ pub trait View {
         layout_rect
     }
 
+    // QUESTION: what's the difference between this and `layout`?
```
matthewgapp, line 186: @jrmoulton do you know the difference between compute_layout and layout?

src/lib.rs
matthewgapp: awwwssome, i've added that

src/view.rs
jrmoulton: I'm not 100% certain actually but the way I've been using them is that the layout is used to update taffy and update the layout of the taffy node. Then the compute_layout gets the taffy layout and then gets the rectangle that is the outer size of the view

src/lib.rs
```diff
@@ -1,3 +1,114 @@
...
+//! Styles are divided into two parts:
+//! [`ComputedStyle`]: A style with definite values for most fields.  
+//!
+//! [`Style`]: A style with [`StyleValue`]s for the fields, where `Unset` falls back to the relevant
+//! field in the [`ComputedStyle`] and `Base` falls back to the underlying [`Style`] or the
+//! [`ComputedStyle`].
+//!
+//!
+//! ## Render loop and update lifecycle
+//! TBD: event -> update -> layout -> paint.
```
matthewgapp, line 95: @dzhou121 or @jrmoulton would love an overview here. I can spend more time going through the code but would love to learn how it's supposed to work at a high level from you.

src/lib.rs
```diff
@@ -1,3 +1,114 @@
...
+//!
+//! potentially discuss update messages, deferred update messages, animations
+//!
+//! ## Notable invariants and tolerances
```
matthewgapp, line 99: Not sure if this section is all that useful. What I'm trying to do is provide some context/guarantees to build up a developer's mental model of how things are supposed to work. The browser has a bunch of these invariants (like one element can be focused at a time, there is one root element (except for in the case of an iframe), events bubble up from their source through all parents to the document, etc. 

I'd really like to build up a Floem mental model for the developer. What fundamental things are different from a browser? Event propagation, focus state, tree shape, etc.

src/app_handle.rs
```diff
@@ -16,33 +16,30 @@ use crate::{
...
-    handle: glazier::WindowHandle,
-    app_state: AppState,
-    paint_state: PaintState,
-
-    file_dialogs: FileDialogs,
-}
-
+// Primarily used to mint and assign a unique ID to each view.
+// It also contains a constant scope which is used to create signals.
+// QUESTION: would it be better to call this `TreeContext`? Since it is really only about minting Ids given the current position in the tree.
```
matthewgapp, line 39: Would it be better to call this `TreeContext`? Since it is really only about minting Ids given the current position in the tree.

src/lib.rs
```diff
@@ -1,3 +1,114 @@
+//! # Floem
+//! Floem is cross-platform GUI framework for Rust 🦀. It aims to be extremely performant while providing world-class developer ergonomics.
+//!
+//! ## Views
+//! Floem models the UI using a tree of [Views](view::View) that are built once. `Views` react to state changes and events.
+//! Views ar self-contained components that can be composed together to create complex UIs.
+//!
+//! ## Ids and Id paths
```
jrmoulton, line 8: I don't think most developers using floem are going to need to understand ids. Especially not at first. I personally wouldn't add ID's to the top module documentation

src/lib.rs
```diff
@@ -1,3 +1,114 @@
...
+//! Events are passed down from the window to the view tree. Each view can handle events and propagate them to its children.
+//! Events can be passed to a specific view using its [Id](id::Id) path. [Id](id::Id) paths are unique ways to quickly identify a view, its children and its parent in the tree.
+//!
+//! Floem is responsible for making a view active. for handling drag and drop mechanics
+//! It's a View's responsibility to handle events and decide whether they're handled.
+//!
+//! ## State management  
+//!
+//! You will probably want your view components to have some state. You should place any state that affects
+//! you view inside a signal so that you can react to updates and update the `View`. Signals are reactive values that can be read from and written to.
```
jrmoulton, line 21: I'd add a link to the excellent leptos_reactive documentation here and point people there (and also maybe to the leptos book) to learn more about reactivity / signals.

src/lib.rs
```diff
@@ -1,3 +1,114 @@
...
+//!
+//! ### Use state to update your view
+//!
+//! To affect the layout and rendering of your component, you will need to send a state update to your component with [Id::update_state](id::Id::update_state)
```
jrmoulton, line 25: I think in general most time (especially for anything not super complex) developing floem will be done by combining pre-existing views provided by floem and not creating completely new ones. I think the top level documentation should reflect that and focus on composing views. I think that is a more user friendly introduction as well.

matthewgapp: Cool, I will rearrange to reflect a higher-level approach. I think this "how to build views" stuff could fit nicely at the top of `view.rs` with an "Advanced" header/section here that links to that.

dzhou121:
Thanks so much for this PR. This is invaluable work, because docs are probably as important as the code itself for users!

As for hosting part, I guess we can use docs.rs once we land ourselves to crates.io?

src/app_handle.rs
dzhou121: Probably we can call it ```ViewContext```? since they are the ```scope``` and ```id``` for a view.

src/context.rs
dzhou121: Maybe ```Deref``` ```LayoutCx``` to ```AppState```?

src/context.rs
```diff
@@ -674,6 +684,7 @@ impl<'a> LayoutCx<'a> {
         self.saved_window_origins.clear();
     }
 
+    // QUESTION: should this be private?
```
dzhou121, line 687: They could be useful for custom Views I guess.

src/lib.rs
dzhou121: Users probably only need to know about ```Style```. ```ComputedStyle``` is more an internal thing.

src/lib.rs
dzhou121: Pointer events will only be sent to the ```active``` View. It is useful when you drag things, e.g. the scroll bar, you set the scroll bar active after pointer down, then when you drag, the ```PointerMove``` will always be sent to the View, even if your mouse is outside of the view. 

And ```focused``` is for sending keyboards events. Keyboard events are only sent to focused view.

src/lib.rs
dzhou121: The cycle is correct.

After an event comes in (e.g. the user clicked the mouse, pressed a key etc), the event will be propagated from the root view to the children. The parent will decide whether sending the event to the child(ren) based on the logic in the ```event``` method in the parent View. There's also event listeners that users can use to respond to events the users choose. The event propagation is stopped whenever a child or an event listener returns ```false``` on the event handling. 

During the event handling, there could be state changes through the reactive system. E.g., on the counter example, when you click increment, it updates the counter and because the label listens to the change (probably a link to the doc of the reactive system?), the label will update the text it presents. The reactive system can't directly manipulate the state of the label because the ```AppState``` owns all the views. And instead, it will send the update to a message queue. After the event propagation is done, Floem will process all the update messages in the queue, and it can manipulate the state of a particular view through the ```update``` method. 

The update of states on the Views could cause some of them need a new layout recalculation, because the size might have changed etc. The ```layout``` method is called from the root view to re-layout the views that have requested a layout call. The layout call is to change the layout properties at ```Taffy```, so after the layout call is done, ```compute_layout``` is called to calculate the sizes and positions of each view. 

And in the end, ```paint``` is called to render all the views to the screen.

src/app_handle.rs
matthewgapp: good call

src/context.rs
matthewgapp: I like that, I'll do that in a separate PR.

src/lib.rs
matthewgapp: Amazing description! Thank you

src/context.rs
matthewgapp: I made it public. @dzhou121 to confirm

src/lib.rs
matthewgapp: I removed reference to computed style in lib.rs

src/view.rs
matthewgapp: I think this is kinda answered now based on what I discerned from @dzhou121's explanation of the render lifecycle.

matthewgapp:
> Thanks so much for this PR. This is invaluable work, because docs are probably as important as the code itself for users!
> 
> As for hosting part, I guess we can use docs.rs once we land ourselves to crates.io?

creates.io would work for hosting! I suggested GH pages to host docs in the meantime because I'm not sure how long until this will be published on crates.io

src/lib.rs
matthewgapp: done!

dzhou121:
Great work! Thanks.

PR #61, opened 2023-06-11
Rename AppContext to ViewContext

matthewgapp:
AppContext doesn't really have anything to do with the App. It's relative to the view and primarily for minting and managing Ids. We should rename it to ViewContext (per [this conversation](https://github.com/lapce/floem/pull/60#discussion_r1225790875)) 

It helps calls like the one below read much better. 

```rust
 let (id, children) = ViewContext::new_id_with_child(children);
 ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ A native Rust UI library with fine-grained reactivity
 It's still early days so expect lots of things missing!
 
 ```rust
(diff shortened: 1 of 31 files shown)


matthewgapp:
Tested all examples to make sure they compile

PR #63, opened 2023-06-11
Expose app_state on layout context 

matthewgapp:
Custom Views like scroll, tab, text_input and more need access to the app state during layout. Therefore, we should allow View authors access to the app state in the `layout` method via the `LayoutCx` 

It was suggested in [this comment](https://github.com/lapce/floem/pull/60/files#r1225791505) that we should implement Deref. I've reconsidered after some though and I think we should make it an explicit access to keep the mental model clean for the developer. Deref can often feel and look like inheritence and come with some of the downsides, like obscuring which methods belong and affect which data. 

Leaving the API explicit allows us to implement Deref in the future without a breaking change. If we implemented Deref now, we'd have to break people's APIs if we decided to undo that at a later point.

### DIFF
--- a/src/app_handle.rs
+++ b/src/app_handle.rs
@@ -233,27 +233,10 @@ impl<V: View> AppHandle<V> {
     }
 
     fn layout(&mut self) {
-        let mut cx = LayoutCx {
-            app_state: &mut self.app_state,
-            viewport: None,
-            color: None,
-            font_size: None,
-            font_family: None,
-            font_weight: None,
-            font_style: None,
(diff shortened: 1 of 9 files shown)


src/app_handle.rs
```diff
@@ -233,27 +233,10 @@ impl<V: View> AppHandle<V> {
     }
 
     fn layout(&mut self) {
-        let mut cx = LayoutCx {
```
matthewgapp, line 236: moved this to a constructor because the app_state field is now private.

Issue #66, opened 2023-06-12
Support views with single child dynamically updating


jrmoulton:
Child views should update dynamically in response to signal changes. 

Eg.
```rust
fn wave_options(wave_type: ReadSignal<WaveType>) -> impl View {
    container_box(move || match wave_type.get() {
        WaveType::WaveSelect => Box::new(default_wave_options()),
        WaveType::CustomFunction => Box::new(custom_function()),
    })
    .style(|| Style::BASE.flex_col().width_pct(100.))
}
```

One thing to note though is that because the views are strongly typed any place where the child changes dynamically the possible children must all be of the exact same view type (which includes all children). Using BoxContainer would be necessary in any cases where the children weren't exactly the same. 

I tried implementing this on BoxContainer but kept running into issues

```rust
pub fn container_box(child: impl Fn() -> Box<dyn View> + 'static) -> ContainerBox {
    let cx = ViewContext::get_current();
    let id = cx.new_id();
    let mut curr_cx = cx;
    curr_cx.id = id;
    ViewContext::save();
    ViewContext::set_current(curr_cx);
    create_effect(cx.scope, move |_| {
        let child_id = curr_cx.new_id();
        let mut child_cx = curr_cx;
        child_cx.id = child_id;
        ViewContext::save();
        ViewContext::set_current(child_cx);
        let new_child = child();
        id.update_state(new_child, false);
        ViewContext::restore();
    });
    ViewContext::restore();
    let (_temp_child_id, temp_child) = ViewContext::new_id_with_child(crate::views::empty);
    ContainerBox {
        id,
        child: Box::new(temp_child),
    }
}
```

This is probably not the right approach. I run into issues with the ids. Inside of the effect I always get a mutable borrow error when creating the new id.
https://github.com/lapce/floem/blob/a0f29c6965f9d6bfe220cb410f0e58d2ed330250/src/id.rs#L68

Is there a different way I should be approaching this?

jrmoulton:
> It's probably like this

For me this still panics in the same place with a borrow_mut error

dzhou121:
It doesn't for me. Maybe paste the full trace?

dzhou121:
It's because these two has already borrowed it. 

https://github.com/lapce/floem/blob/main/src/app_handle.rs#L654
https://github.com/lapce/floem/blob/main/src/app_handle.rs#L701

You'll need to get the id_path out like here https://github.com/lapce/floem/blob/main/src/app_handle.rs#L472

[...]

PR #67, opened 2023-06-12
Add reactive child to container box

jrmoulton:
Sorry. Probably should've opened a PR to start with

Previous conversation in #66

### DIFF
--- a/src/views/dyn_container.rs
+++ b/src/views/dyn_container.rs
@@ -1,0 +1,120 @@
+use std::marker::PhantomData;
+
+use floem_reactive::create_effect;
+use glazier::kurbo::Rect;
+
+use crate::{
+    app_handle::ViewContext,
+    id::Id,
+    view::{ChangeFlags, View},
+};
+
+pub struct DynamicContainer<CF: Fn(T) -> Box<dyn View> + 'static, T: 'static> {
+    id: Id,
+    child: Box<dyn View>,
+    child_fn: CF,
+    phantom: PhantomData<T>,
+    cx: ViewContext,
+}
+
+pub fn dyn_container<CF: Fn(T) -> Box<dyn View> + 'static, T: 'static>(
+    update_view: impl Fn() -> T + 'static,
+    child: CF,
+) -> DynamicContainer<CF, T> {
+    let cx = ViewContext::get_current();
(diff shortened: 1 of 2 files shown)


dzhou121:
Is it only the new views not update correctly or the original view as well?

jrmoulton:
The original view (the container_box) seems to update just fine and the layout for it and the child is correct. But for example if I have a label somewhere in the child view the text does not appear

jrmoulton:
https://github.com/lapce/floem/assets/13090441/150ed8fb-9e95-4ea5-a458-ef27cf5641a5

When there are events that cause an update then the text appears

dzhou121:
I think I know what the issue now. 

When calling ```child()```, they would generate state updates, and these will be before the state change of the container box. So when the child view state update dispatches, they are not with the container box yet, so nothing happens. 

There's another paper cut. Every signal will trigger a new generated View. So it's very easy to cause unwanted view changes.

So we'll need an approach similar to list, an additional closure to trigger the view update. And the view creation needs to happen in ```update````

jrmoulton:
Ahh interesting. 
https://github.com/lapce/floem/blob/7bcd3725290765f9d69374b883a13283646daac8/src/views/list.rs#L383C5-L395C12

I'm assuming this is the bit of the code that is most relevant?

jrmoulton:
I've moved the view creation to the update function but I'm still having the same problem except now clicking doesn't cause any updates on the child. I'm pretty lost at this point and I definitely don't understand how this works completely. 

@dzhou121

jrmoulton:
I think 6feda1f isn't far from correct. I think children just aren't getting the correct context. When I add debug statements to the elements every function does get run to build all of the views but they just aren't showing up.

jrmoulton:
And from my debug statements it doesn't look like the view is being rebuilt on every signal change in the view. Only the one that should cause the view to change so it looks that that 6feda1f does have the correct behavior in response to signal changes

jrmoulton:
Is there any way that the saving / restoring of layout and paint contexts gets messed up or doesn't account for views being built out of order because they are getting built reactively @dzhou121 ?

jrmoulton:
🎉 🎉 🎉

PR #80, opened 2023-08-01
Added right click event handler.

Wcubed:
Added an `on_right_click` event handler, similar to the one for `on_click`.
The event fires when the right click button is released.

Updated the widget gallery example with a button that activates on right-clicking.


Wcubed:
That is a very good point. I've changed the name of the new event handler to use "secondary" instead of "right". Question is indeed if we'd want to extend that through the rest of floem.

waywardmonkeys:
>  im not sure if its completed

It is on glazier's `main` branch. I did it. (Along with some other clean ups to pointer events.

PR #89, opened 2023-08-19
Update docs

jrmoulton:
This fixes several issues with warnings in the docs for public documentation linking to private types and for linking to non-existing items. This also updates and adds to the Floem top module documentation as well as a few other additions


jrmoulton:
Would it be useful to have the CI test the documentation as well?

jrmoulton:
> Would it be useful to have the CI test the documentation as well?

Ah nevermind. Looks like it already does

PR #92, opened 2023-08-20
Add docs for {base_,}style Decorators

MinusGix:
I kept putting this off due to wanting to write more documentation past them, but I decided to just PR these rather than wait until I got around to that.  
This just adds some basic example documentation to `style`/`base_style`

### DIFF
--- a/src/views/decorator.rs
+++ b/src/views/decorator.rs
@@ -13,6 +13,28 @@ use crate::{
 };
 
 pub trait Decorators: View + Sized {
+    /// Alter the style of the view.  
+    ///
+    /// -----
+    ///
+    /// Note: repeated applications of `style` overwrite previous styles.  
+    /// ```rust
+    /// # use floem::{peniko::Color, view::View, views::{Decorators, label, stack}};
+    /// fn view() -> impl View {
+    ///     label(|| "Hello".to_string())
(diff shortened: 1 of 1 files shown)


src/views/decorator.rs
```diff
@@ -13,6 +13,28 @@ use crate::{
...
+    /// -----
+    ///
+    /// Note: repeated applications of `style` overwrite previous styles.  
+    /// If you are returning from a function that produces a view, you may want
+    /// to use `base_style` for the returned [`View`] instead.  
+    /// ```rust
+    /// # use floem::{peniko::Color, view::View, views::{Decorators, label, stack}};
+    /// fn view() -> impl View {
+    ///     label(|| "Hello".to_string())
+    ///         .style(|| Style::default().font_size(20.0).color(Color::RED))
```
dzhou121, line 27: should be `base_style` here?

MinusGix: That's intended to show how `style` behaves if you use it on a view that's returned, to give an illustration of what goes wrong.
(Though I could see that as potentially being confusing, so perhaps it should just be the note?)

dzhou121: Ah got you. 

Yeah maybe move "If you are returning from a function that produces a view, you may want to use `base_style` for the returned [`View`] instead. " under the code example.

PR #93, opened 2023-08-22
Disable the wayland feature by default

DJMcNab:
But expose a feature to re-enable it.

This frees us to enable the wayland feature by default in Glazier, to ease development there, whilst still giving it breathing room


dzhou121:
Thanks!

PR #96, opened 2023-08-24
Add round scroll bar option

jrmoulton:
This adds the option for a scroll bar to be rounded and makes that the default on macOS

Update: this now also 
1. changes the View style methods to take a base style in function argument
2. Moves the scroll bar styles to the View style and makes them apply to all children elements until override like font size

### DIFF
--- a/src/app_handle.rs
+++ b/src/app_handle.rs
@@ -302,6 +302,15 @@ impl<V: View> AppHandle<V> {
             saved_font_styles: Vec::new(),
             saved_line_heights: Vec::new(),
             saved_z_indexes: Vec::new(),
+            scroll_bar_color: None,
+            scroll_bar_rounded: None,
+            scroll_bar_thickness: None,
+            scroll_bar_edge_width: None,
+
+            saved_scroll_bar_colors: Vec::new(),
+            saved_scroll_bar_roundeds: Vec::new(),
+            saved_scroll_bar_thicknesses: Vec::new(),
+            saved_scroll_bar_edge_widths: Vec::new(),
         };
         cx.paint_state.renderer.as_mut().unwrap().begin();
         self.view.paint_main(&mut cx);
--- a/src/context.rs
+++ b/src/context.rs
@@ -681,6 +681,10 @@ pub struct LayoutCx<'a> {
     app_state: &'a mut AppState,
     pub(crate) viewport: Option<Rect>,
     pub(crate) color: Option<Color>,
+    pub(crate) scroll_bar_color: Option<Color>,
+    pub(crate) scroll_bar_rounded: Option<bool>,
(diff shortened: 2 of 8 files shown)


dzhou121:
Maybe let's create a `ScrollBarStyle` which has bar color, bar size, and radius etc.

jrmoulton:
I think this is good. I don't love having to import ScrollBarStyle and create a new instance of it to add it to the function. Something SwiftUI-esque like
```rust
.scroll_bar_style( .color(Color::GREEN) )
```

would be nice but I don't know of any way to do this so I think this is good enough

jrmoulton:
I guess this is possible...

```rust
    pub fn scroll_bar_style(
        self,
        style: impl Fn(ScrollBarStyle) -> ScrollBarStyle + 'static,
    ) -> Self {
        let id = self.id;
        create_effect(move |_| {
            let style = style(ScrollBarStyle::BASE);
            id.update_state(UpdateState::Style(style), false);
        });

        self
    }
```
and then
```rust
        .scroll_bar_style(|base| base.color(custom_scroll_bar_color()))
```

jrmoulton:
> and then
> 
> ```rust
>         .scroll_bar_style(|base| base.color(custom_scroll_bar_color()))
> ```

I actually really quite like this for styles. Would actually be a nice way to do View styles as well

jrmoulton:
#97 affects how this is viewed

dzhou121:
I agree. This is very nice. We could do the same for view style.

dzhou121:
On the scroll bar style, I feel like we need to put it in the view style and let it affect child views like the way font size does. So that you can only specify it once at the root. Still you have the ability to customise it at a child view.

jrmoulton:
> On the scroll bar style, I feel like we need to put it in the view style and let it affect child views like the way font size does. So that you can only specify it once at the root. Still you have the ability to customise it at a child view.

Ok. Looks like that's done by adding a current / saved value to the paint and a layout contexts?

dzhou121:
> > On the scroll bar style, I feel like we need to put it in the view style and let it affect child views like the way font size does. So that you can only specify it once at the root. Still you have the ability to customise it at a child view.
> 
> Ok. Looks like that's done by adding a current / saved value to the paint and a layout contexts?

Yes.

jrmoulton:
And when adding these to the view style should we keep them together (in a ScrollBarStyle) struct or separate them? Together seems better for the implementation but it'll be ugly to use

dzhou121:
Feels like we should seperate them since all other styles are "flat".

jrmoulton:
Minimal testing but does seem to work with applying to child views and overrides also work.This also switches the `style` method to take a Style and the BASE style is passed in

PR #116, opened 2023-10-08
Add support for images

presiyan-ivanov:
image


panekj:
What's the source for the images?

presiyan-ivanov:
https://rustacean.net/ for Ferris and https://www.pexels.com/photo/sunflower-selective-focus-photography-1366630/ for the sunflower. Both do **not** require attribution and are free to use

PR #117, opened 2023-10-08
fix unwrap crash when render surface is unavailable (0 height etc)

Keshi:
### Repro
1.  Run examples/responsive (or probably any example?)
2.  Vertically resize the window to the smallest possible so that only menu bars are showing (0 height surface)

### Observed
- Crash
- surface.get_current_texture() returns Err

This PR makes the UI just temporarily not render, instead of crashing the application.


dzhou121:
Thanks!

PR #118, opened 2023-10-10
Add a `tiny_skia` rendering backend

Zoxc:
This adds a `tiny_skia` rendering backend which is used if `vger` fails to initialize or if `wgpu` uses a CPU backend.

It doesn't support blur radios and Z-indexing. Masking is currently quite expensive as `tiny_skia` doesn't support masking by rectangles.


dzhou121:
Wow thanks so much! That's blazing fast for you to add this feature.

PR #124, opened 2023-10-26
Extend Winit window configuration, change default rendering colour

panekj:
…parent

### DIFF
--- a/vger/src/lib.rs
+++ b/vger/src/lib.rs
@@ -446,7 +446,7 @@ impl Renderer for VgerRenderer {
                             r: 0.0,
                             g: 0.0,
                             b: 0.0,
-                            a: 0.5,
+                            a: 0.0,
                         }),
                         store: StoreOp::Store,
                     },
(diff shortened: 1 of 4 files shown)


vger/src/lib.rs
```diff
@@ -442,7 +442,12 @@ impl Renderer for VgerRenderer {
                     view: &texture_view,
                     resolve_target: None,
                     ops: wgpu::Operations {
-                        load: wgpu::LoadOp::Clear(wgpu::Color::WHITE),
+                        load: wgpu::LoadOp::Clear(wgpu::Color {
```
dzhou121, line 445: maybe only do this if the window is transparent? also why a is 0.5 instead of 0.0?

panekj: Is there such configuration? The transparency setting is for winit to signal to OS that it might be transparent.

> also why a is 0.5 instead of 0.0?

my own custom thing, will fix in a bit

dzhou121: Yeah that's what I meant, when the WindowConfig has transparent setting in it.

panekj: I have no idea how I'm supposed to do that, winit doesn't even expose the transparency state in Window, only in WindowBuilder

PR #127, opened 2023-10-30
Add a `hover` style selector

Zoxc:
This adds a `hover` style selector to the `Style` type. It can be used like:
```rust
style(|s| s.hover(|s| s.scroll_bar_color(Color::ROYAL_BLUE)));
```
Currently it only works with the properties defined by the `prop!` macro.


jrmoulton:
This is the API I want for all hover/modifier styles. This would make it really easy to make a single style that can be applied to a view that would be fully reactive. Reactivity using some predefined colors/styles would become beautifully trivial.

PR #128, opened 2023-10-31
Use the `prop!` macro for all style properties

Zoxc:
This replaces the builtin style field properties with properties defined by the `proc!` macro. This makes them work with the `hover` method and animations (for some types).

This is based on top of https://github.com/lapce/floem/pull/127.

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -25,49 +25,69 @@
 //! ```
 //!
 
-use floem_renderer::cosmic_text::{LineHeightValue, Style as FontStyle, Weight};
+use floem_renderer::cosmic_text;
+use floem_renderer::cosmic_text::{LineHeightValue, Weight};
 use peniko::Color;
-use std::any::Any;
+use std::any::{type_name, Any};
 use std::collections::hash_map::Entry;
 use std::collections::HashMap;
+use std::fmt::{self, Debug};
 use std::hash::Hash;
 use std::hash::Hasher;
 use std::ptr;
 use std::rc::Rc;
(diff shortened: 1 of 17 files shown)


src/style.rs
```diff
@@ -315,211 +474,205 @@ impl<T> From<T> for StyleValue<T> {
...
+            .get_prop::<P>()
+            .unwrap_or_else(|| P::default_value())
+    }
+
+    pub(crate) fn get_builtin(&self) -> BuiltinStyleReader<'_> {
+        BuiltinStyleReader { style: self }
+    }
+}
+#[derive(Debug, Clone)]
+pub struct Style {
```
dzhou121, line 494: I'm wondering if it's better to move StyleMap content here. e.g.

#[derive(Default, Clone)]
pub struct Style {
    pub(crate) map: HashMap<StylePropRef, StyleMapValue<Rc<dyn Any>>>,
    pub(crate) selectors: HashMap<StyleSelector, StyleMap>,
}

Zoxc: There's quite a bit of stuff that can be refactored with this PR. I feel like the PR is sufficiently large already though.

dzhou121: Yeah we can merge this first and work on stuff in another PR.

dzhou121:
don't know why this has conflicts after merging #127

Zoxc:
GitHub's modifying the commits when you merge, so it didn't skip the commit for https://github.com/lapce/floem/pull/127.

PR #134, opened 2023-11-01
Add a view inspector

Zoxc:
This adds an inspector window for views which lets you view the view tree and layout of views. It only does render captures with the `vger` renderer.

It's hooked up to the widget gallery and counter example (press F11 to open it). It's currently quite slow so I recommend testing it on the counter example.

### DIFF
--- a/src/inspector.rs
+++ b/src/inspector.rs
@@ -1,0 +1,457 @@
+use crate::app::{add_app_update_event, AppUpdateEvent};
+use crate::context::AppState;
+use crate::event::{Event, EventListener};
+use crate::id::Id;
+use crate::new_window;
+use crate::view::View;
+use crate::views::{
+    dyn_container, empty, img_dynamic, list, scroll, stack, text, Decorators, Label,
+};
+use crate::window::WindowConfig;
+use floem_reactive::{create_rw_signal, RwSignal, Scope};
+use image::DynamicImage;
+use kurbo::{Point, Rect, Size};
+use peniko::Color;
+use std::cell::Cell;
+use std::fmt::Display;
+use std::rc::Rc;
+use std::time::Instant;
+use taffy::style::AlignItems;
+use winit::window::WindowId;
+
+#[derive(Clone, Debug)]
+pub struct CapturedView {
+    id: Id,
+    name: String,
+    layout: Rect,
+    clipped: Rect,
+    children: Vec<Rc<CapturedView>>,
+}
+
+impl CapturedView {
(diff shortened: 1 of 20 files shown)


dzhou121:
Impressive work!

If I understand it correctly, the capture is to provide a snapshot of the whole UI so that you can give the area when you chose a view in the view tree.

Would it better to have an overlay on the original root view (just like the bit I did for Linux context menu), so that you can directly highlight the area of a view on the original UI, without the need to do a "screenshot", also it's more "real time".

src/inspector.rs
```diff
@@ -0,0 +1,457 @@
...
+        move || selected.get(),
+        move |current| {
+            let info = |i| text(i).style(|s| s.padding(5.0));
+            if let Some(view) = current.and_then(|id| capture.root.find(id)) {
+                let name = info(format!("Type: {}", view.name));
+                let count = info(format!("Child Count: {}", view.children.len()));
+                let x = info(format!("X: {}", view.layout.x0));
+                let y = info(format!("Y: {}", view.layout.y0));
+                let w = info(format!("Width: {}", view.layout.width()));
+                let h = info(format!("Width: {}", view.layout.height()));
```
presiyan-ivanov, line 258:
```suggestion
                let h = info(format!("Height: {}", view.layout.height()));
```

Zoxc:
I intentionally went with an approach with immutable UI snapshots. This keeps the inspector simple and it's helpful for debugging temporal issues. It could be extended to capture multiple frames for animation debugging.

Having an overlay on the real UI with this approach could easily get out of sync so I think that could be confusing or weird. It's also a bit more complex.

dzhou121:
I was thinking of a browser inspector like of a tool.

Zoxc:
I added a `get_content_rect` method to `PaintCx` and used it for `img`. It's used to get the rect inside padding and borders when painting.

dzhou121:
It crashes on my mac due to wrong TextureFormat

```
Incompatible color attachments at indices [0]: the RenderPass uses textures with formats [Some(Rgba8Unorm)] but the RenderPipeline uses attachments with formats [Some(Bgra8Unorm)]
```
So you might need to use the texture_format that's found.

dzhou121:
also window scaling isn't applied by the looks of it. 

Screenshot 2023-11-01 at 21 34 11

Zoxc:
I think I have those issues fixed now.

Zoxc:
I added inspection of view styles too (with colors).

PR #136, opened 2023-11-02
Add style classes

Zoxc:
This adds style classes. They can be declared with the `style_class!` macro:
```rust
style_class!(pub Button);
```
There is a view decorator to apply them:
```rust
view.class(Button)
```
And a method on `Style` to apply styles only to a class:
```rust
s.class(Button, |s| s.border(1.0))
```

The `scroll` view now uses classes for the handle and track styles. The window scale example is updated to use a button class.


dzhou121:
This is really cool!

PR #142, opened 2023-11-04
Add a style pass

Zoxc:
This adds a style pass to allow layout to be skipped when unnecessary. It will automatically trigger layout if the styling used by Taffy is changed. It makes the performance of the view inspector decent as triggering a full Taffy layout is rarer.

It adds `request_style`, `request_paint` and `request_all`. The last one is useful when adding new views to the tree. Existing calls to `request_layout` for Floem users should be re-evaluated as `request_layout` no longer computes styling. Reading of styles is moved to `View::style`.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -51,8 +51,9 @@ pub struct ViewState {
     pub(crate) node: Node,
     pub(crate) children_nodes: Vec<Node>,
     pub(crate) request_layout: bool,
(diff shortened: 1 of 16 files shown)


dzhou121:
Just found there seems to be a bit of issue. Please see the incorrect horizontal scroll on the left tab switcher. 

Screenshot 2023-11-05 at 15 48 33

PR #143, opened 2023-11-05
Add style property transitions

Zoxc:
This add style property transitions which can be used by the `transition` method on `Style`. For example: `transition(Background, Transition::linear(0.1))`. Users of `prop_extracter!` will get these animations applied if the property type supports them (current `f64` and `Color`).

The `themes` example is updated to use them.


presiyan-ivanov:
LGTM

presiyan-ivanov (approved):

PR #147, opened 2023-11-05
Add defaults for `View` trait methods

Zoxc:
This adds default methods to the `View` trait which generally work with views with or without children. This is quite helpful in reducing code duplication across view impls. It also serves as a reference for new view implementations.

`for_each_child` and `for_each_child_mut` are added to `View` to iterate over children without allocation. The `children` and `children_mut` methods can probably be removed, but I haven't looked carefully at their uses yet.

### DIFF
--- a/src/view.rs
+++ b/src/view.rs
@@ -111,6 +111,14 @@ bitflags! {
 pub trait View {
     fn id(&self) -> Id;
 
+    /// This method walk over children and must be implemented if the view has any children.
+    /// It should return children back to front and should stop if `_for_each` returns `true`.
+    fn for_each_child<'a>(&'a self, _for_each: &mut dyn FnMut(&'a dyn View) -> bool) {}
+
+    /// This method walk over children and must be implemented if the view has any children.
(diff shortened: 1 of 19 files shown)


src/view.rs
```diff
@@ -152,23 +198,52 @@ pub trait View {
...
     }
 
     /// Implement this to handle events and to pass them down to children
     ///
     /// Return true to stop the event from propagating to other views
-    fn event(&mut self, cx: &mut EventCx, id_path: Option<&[Id]>, event: Event) -> bool;
+    fn event(&mut self, cx: &mut EventCx, id_path: Option<&[Id]>, event: Event) -> bool {
+        let mut handled = false;
+        self.for_each_child_mut(&mut |child| {
+            handled |= cx.view_event(child, id_path, event.clone());
```
dzhou121, line 233: We'll probably break out if the event is handled.

Zoxc: Returning true breaks the loop here.

dzhou121: ah right

PR #151, opened 2023-11-06
Refactor `ChangeFlags` and requests

Zoxc:
This refactors the use of `ChangeFlags`. `ViewState` has a new `requested_changes: ChangeFlags` field which replaces `request_layout` and `request_style`. `View::update` on longer returns `ChangeFlags` as view implementation are expected to call `request_*` functions.

### DIFF
--- a/src/window_handle.rs
+++ b/src/window_handle.rs
@@ -22,7 +22,7 @@ use crate::views::{container_box, stack, Decorators};
 use crate::{
     animate::{AnimPropKind, AnimUpdateMsg, AnimValue, AnimatedProp, SizeUnit},
     context::{
-        AppState, ChangeFlags, EventCx, LayoutCx, MoveListener, PaintCx, PaintState,
+        AppState, ChangeFlags, EventCx, FrameUpdate, LayoutCx, MoveListener, PaintCx, PaintState,
         ResizeListener, StyleCx, UpdateCx,
     },
     event::{Event, EventListener},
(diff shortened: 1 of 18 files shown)


src/window_handle.rs
```diff
@@ -624,22 +625,22 @@ impl WindowHandle {
...
-            flags |= self.process_deferred_update_messages();
-            flags |= self.process_anim_update_messages();
+            self.process_deferred_update_messages();
+            self.process_anim_update_messages();
         }
 
         self.set_cursor();
 
-        if !flags.is_empty() {
+        if paint {
```
dzhou121, line 643: What if a view just wanted to request_paint without style or layout change, e.g. a canvas?

Zoxc: They can just call `request_paint` to do that.

dzhou121: Ah I see it ```request_style``` now. But is it efficient enough when the canvas wants some animation frames.

Zoxc: It only requests style on one view, but I'm going to replace it to just requesting painting.

Zoxc:
I've added some further changes here. `schedule_style`, `schedule_layout` and `schedule_paint` are new functions which request a pass on the *next* frame. These are now used to drive animations.

Timers are no longer used for animations, which makes the `themes` example have smooth animations.

The `animation` example now works again.

Paint requests no longer invokes the style pass.

I've made `ChangeFlags` private for consistency as painting needs damage tracking which doesn't have a tree structure.

PR #154, opened 2023-11-07
Add toggle button

jrmoulton:
This works but could definitely be cleaned up. @Zoxc this is the first time I'm using the style classes/props and I'd appreciate any feedback. Especially if I'm missing the point. 

There are sizing-related things that are set inside of `paint` which feels weird to me but I couldn't figure out where else to access the properties. 

A styled example. 
image


dzhou121:
Looks beautiful!

One thing though is maybe we want to use an event listener style for the toggle value (```on_change``` or ```on_toggle```) rather than a signal. (And change checkbox and text input to use event listener style as well, but that can be in a different PR)

src/views/toggle_button.rs
```diff
@@ -0,0 +1,151 @@
...
+use kurbo::{Point, Size};
+use winit::keyboard::{Key, NamedKey};
+
+use crate::{id, prop, style_class, view::View};
+
+use super::Decorators;
+
+prop!(pub ToggleBg: Option<peniko::Color> {} = None);
+prop!(pub ToggleFg: Option<peniko::Color> {} = None);
+style_class!(pub ToggleClass);
```
Zoxc, line 12: `ToggleButtonClass` to match?

src/views/toggle_button.rs
```diff
@@ -0,0 +1,151 @@
...
+use floem_renderer::Renderer;
+use kurbo::{Point, Size};
+use winit::keyboard::{Key, NamedKey};
+
+use crate::{id, prop, style_class, view::View};
+
+use super::Decorators;
+
+prop!(pub ToggleBg: Option<peniko::Color> {} = None);
+prop!(pub ToggleFg: Option<peniko::Color> {} = None);
```
Zoxc, line 11: I'd add a builtin `Foreground foreground nocb: Option<Color> {} = None,` property and use that along with `Background`. They should be read in the `style` pass, see `label.rs`.

src/views/toggle_button.rs
```diff
@@ -0,0 +1,151 @@
...
+            _ => {}
+        };
+        false
+    }
+
+    fn paint(&mut self, cx: &mut crate::context::PaintCx) {
+        let layout = cx.get_layout(self.id).unwrap();
+        let size = Size::new(layout.size.width as f64, layout.size.height as f64);
+
+        self.width = size.width;
```
Zoxc, line 98: This seems like it belongs in `compute_layout`.

src/views/toggle_button.rs
jrmoulton: The issue with using `Background` is that painting the background is a special case that is handled in `paint_view`. And that would require that the user always set a border radius. Which could be done in a default style?

jrmoulton: But that border radius needs be relative to the size of the element which isn't super convenient to set as a user

jrmoulton:
Added custom inset and size properties

image

jrmoulton:
Should be good to go

jrmoulton:
I've switched the button to use the regular background and added a built-in foreground style. I've also made it so that the border radius takes a PxPct so that a percentage border radius can be set.

PR #189, opened 2023-11-19
Add a tooltip view and widget

Zoxc:


Zoxc:
For correct styling, https://github.com/lapce/vger-rs/pull/4 or https://github.com/lapce/vger-rs/pull/3 should be merged.

PR #203, opened 2023-11-25
Use copypasta for clipboard handling

lukaslihotzki:
For correct clipboard handling (fixing https://github.com/lapce/lapce/issues/2787, https://github.com/lapce/lapce/issues/2768, https://github.com/lapce/lapce/issues/2676), the `ClipboardContext` needs to be created once, passing the `RawDisplayHandle` in Wayland. Therefore, it makes sense to do it in floem and offer the ClipboardContext to apps.

### DIFF
--- a/src/clipboard.rs
+++ b/src/clipboard.rs
@@ -1,0 +1,74 @@
+use once_cell::sync::Lazy;
+use parking_lot::Mutex;
+use raw_window_handle::RawDisplayHandle;
+
+#[cfg(not(any(target_os = "macos", windows)))]
+use copypasta::{
+    wayland_clipboard,
+    x11_clipboard::{Primary as X11SelectionClipboard, X11ClipboardContext},
+};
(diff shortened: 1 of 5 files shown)


src/clipboard.rs
```diff
@@ -0,0 +1,46 @@
...
+                X11ClipboardContext::<X11SelectionClipboard>::new().unwrap(),
+            )),
+        };
+
+        #[cfg(any(target_os = "macos", windows))]
+        return Self {
+            clipboard: Box::new(ClipboardContext::new().unwrap()),
+            selection: None,
+        };
+    }
```
dzhou121, line 45: maybe add `get_contents` and `set_contents` to `Clipboard` 

so that user can just call `Clipboard::get_contents()` `Clipboard::set_contents(contents)` without needs to import `CLIPBOARD` and lock it.

PR #226, opened 2023-12-21
Splitter widget

dparnell:
Hi, I have created simple `h_splitter` and `v_splitter` widgets and updated the main example to use the `h_splitter` as an example


src/style.rs
```diff
@@ -221,8 +221,8 @@ pub trait StyleProp: Default + Copy + 'static {
     fn prop_ref() -> StylePropRef;
     fn default_value() -> Self::Type;
 }
-
 #[derive(Debug)]
+
 pub struct StylePropInfo {
```
panekj, line 226: Why change that?

dominikwilkowski:
Out of curiosity: what does the splitter do over the stack? Is it just a stack with exactly two children and some sizing?

dparnell:
Yes, all it does it allow the parts to be sized. If the stack had a sizing option that would be really cool. I guess having a splitter widget is a bit more explicit.

Issue #230, opened 2023-12-25
Descendants of a disabled View apply disabled behavior, but not disabled styling


pieterdd:
I observed the following when putting a View into a disabled state:
- Its descendants exhibit disabled behavior, i.e. the buttons in my test no longer respond to clicks (makes sense)
- Its descendants don't look disabled, i.e. they don't apply disabled styling

Here's a live demo:

https://github.com/lapce/floem/assets/1407980/202d8911-2ca3-4fcd-9a63-469ff18837cb

Source code for the demo:

```rust
use floem::kurbo::Size;
use floem::peniko::Color;
use floem::reactive::{create_signal, ReadSignal, WriteSignal};
use floem::style::AlignItems;
use floem::view::View;
use floem::views::{h_stack, label, v_stack, Decorators};
use floem::widgets::{button, checkbox};
use floem::window::WindowConfig;
use floem::EventPropagation;

fn my_button(label: &'static str) -> impl View {
    button(move || label).style(|s| {
        s.background(Color::GREEN)
            .color(Color::WHITE)
            .disabled(|s| s.background(Color::DARK_GRAY))
            .font_size(18.0)
            .padding(5.0)
    })
}

fn spinbox(counter: ReadSignal<i32>, set_counter: WriteSignal<i32>) -> impl View {
    h_stack((
        my_button("-").on_click(move |_| {
            set_counter.update(|value| *value -= 1);
            EventPropagation::Stop
        }),
        label(move || counter.get()),
        my_button("+").on_click(move |_| {
            set_counter.update(|value| *value += 1);
            EventPropagation::Stop
        }),
    ))
    .style(|s| {
        s.align_items(AlignItems::Center)
            .disabled(|s| s.background(Color::DIM_GRAY))
            .gap(10.0, 0.0)
    })
}

fn app_view() -> impl View {
    let (spinbox_enabled, set_spinbox_enabled) = create_signal(true);
    let (counter, set_counter) = create_signal(0);

    v_stack((
        checkbox(spinbox_enabled)
            .on_click_stop(move |_| set_spinbox_enabled.set(!spinbox_enabled.get())),
        spinbox(counter, set_counter).disabled(move || !spinbox_enabled.get()),
    ))
}

fn main() {
    let window_config = WindowConfig::default().size(Size {
        width: 200.0,
        height: 100.0,
    });

    floem::Application::new()
        .window(move |_| app_view(), Some(window_config))
        .run()
}
```

[...]

PR #231, opened 2023-12-27
Propagate disabled state to descendants

pieterdd:
This is a fix for [Descendants of a disabled View apply disabled behavior, but not disabled styling](https://github.com/lapce/floem/issues/230). The code sample in that thread worked correctly for me after the code change:

image

For more context about why propagation of disabled state matters, see [this example](https://github.com/lapce/floem/discussions/228#discussioncomment-7938256). The approach where classes are applied to child components and modified upon disabling of the parent doesn't work in all cases, plus it requires parent components to interfere with child components. Propagating the disabled state seemed like a better separation of concerns.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -384,7 +384,17 @@ impl AppState {
     }
 
     pub fn is_disabled(&self, id: &Id) -> bool {
-        self.disabled.contains(id)
+        if self.disabled.contains(id) {
+            return true;
+        }
+        if let Some(id_path) = id.id_path() {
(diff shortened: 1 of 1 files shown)


dzhou121:
Thanks for the PR. But this approach is quite expensive.

I tried to fix it in e795021bfb28cd15a6d499349e547c435ceb8520, can you please check if that works?

pieterdd:
I'm happy to report that e795021bfb28cd15a6d499349e547c435ceb8520 works correctly in all cases I tested. Thank you for the fix.

PR #234, opened 2023-12-30
Add clear focus decorator

jrmoulton:


src/views/decorator.rs
```diff
@@ -247,6 +247,15 @@ pub trait Decorators: View + Sized {
         self
     }
 
+    fn clear_focus(self, when: impl Fn() + 'static) -> Self {
```
jrmoulton, line 250: Was the best API I could come up with. Open to ideas

PR #249, opened 2024-01-02
Add `list` with selections, `selected` style selectors and `scroll_to`

Zoxc:
This adds a `list` view which support selections with an accompanying `list` widget.

A `selected` style selector is added to style selected list items.

A `scroll_to` method is added to the `View` trait to allow scrolling to a specific view. That's used for keyboard navigation on lists.

The simple list example now uses the list widget.

### DIFF
--- a/examples/widget-gallery/src/lists.rs
+++ b/examples/widget-gallery/src/lists.rs
@@ -10,7 +10,7 @@ use floem::{
         container, label, scroll, stack, virtual_stack, Decorators, VirtualStackDirection,
         VirtualStackItemSize,
     },
-    widgets::checkbox,
+    widgets::{checkbox, list},
     EventPropagation,
 };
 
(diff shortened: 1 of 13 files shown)


examples/widget-gallery/src/lists.rs
```diff
@@ -26,17 +26,9 @@ pub fn virt_list_view() -> impl View {
...
     scroll(
-        virtual_stack(
-            VirtualStackDirection::Vertical,
-            VirtualStackItemSize::Fixed(Box::new(|| 20.0)),
-            move || long_list.get(),
-            move |item| *item,
-            move |item| label(move || item.to_string()).style(|s| s.height(24.0)),
-        )
-        .style(|s| s.flex_col()),
+        list((0..100).map(|i| label(move || i.to_string()).style(|s| s.height(24.0))))
```
jrmoulton, line 30: for conciseness the to_string could be omitted

PR #255, opened 2024-01-04
Add stackable `on_key_down` and `on_key_up` event handlers

Zoxc:
This adds stackable `on_key_down` and `on_key_up` event handlers.


jrmoulton:
Should this be extended to all event handler?
I'm running into the same problem with on_focus on text input
https://github.com/lapce/floem/blob/5b5b3367955afd34c4c7f0fc71b5eeb9b2133b8d/src/views/text_input.rs#L151

Zoxc:
Yeah, we probably want all view decorators to stack in general.

PR #256, opened 2024-01-04
Implement radio button

pieterdd:
There's minor one open issue I might need some help with: the black circle is a little bit off-center. Any guesses why?

I'll also check if a 100% rounded box looks better.

image

### DIFF
--- a/src/widgets/mod.rs
+++ b/src/widgets/mod.rs
@@ -11,6 +11,7 @@ use crate::{
 };
 use peniko::Color;
 use std::rc::Rc;
+use taffy::style::AlignItems;
 
 mod checkbox;
 pub use checkbox::*;
@@ -31,6 +32,9 @@ pub mod slider;
 
 mod button;
 pub use button::*;
+
+mod radio_button;
+pub use radio_button::*;
 
 mod text_input;
 pub use text_input::*;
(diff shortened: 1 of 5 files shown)


src/widgets/mod.rs
```diff
@@ -132,6 +135,41 @@ pub(crate) fn default_theme() -> Theme {
...
 
+    let radio_button_style = Style::new()
+        .width(20.)
+        .height(20.)
+        .background(Color::WHITE)
+        .active(|s| s.background(active_bg_color))
+        .transition(Background, Transition::linear(0.04))
+        .hover(|s| s.background(hover_bg_color))
+        .focus(|s| s.hover(|s| s.background(focus_hover_bg_color)))
+        .apply(border_style.clone())
```
Zoxc, line 146: I'd apply a 100% border radius after here.

pieterdd: Much better, thank you. The alignment issue is still lurking though:

image

Zoxc: Not quite sure about that. You could try using a centered `empty` instead of an SVG.

pieterdd: Interesting, I'll give that a spin.

pieterdd: I've rewritten it to use `empty`, but the alignment is still slightly off. I've used `.align_items(AlignItems::CENTER).justify_content()` to center the dot.

image

pieterdd:
This is the most aligned I've been able to get it:

image

Even if I add a left margin of 1 to the dot, it looks a bit off. Then it looks aligned too far to the right, so this might be a subpixel accuracy issue.

image

I'm still open to suggestions, but this is the best I've been able to make it so far.

dzhou121:
We can merge this first and improve on that later.

PR #258, opened 2024-01-06
Don't round in the vger-rs renderer

Zoxc:
This makes pixel aligned lines render correctly instead of being split across pixels.

We probably need some better rounding scheme.

@dzhou121 Could you take before and after screenshots of the widget gallery on Mac to see the impact there?

### DIFF
--- a/vger/src/lib.rs
+++ b/vger/src/lib.rs
@@ -158,10 +158,7 @@ impl VgerRenderer {
     fn vger_point(&self, point: Point) -> vger::defs::LocalPoint {
         let coeffs = self.transform.as_coeffs();
         let point = point + Vec2::new(coeffs[4], coeffs[5]);
-        vger::defs::LocalPoint::new(
-            (point.x * self.scale).round() as f32,
-            (point.y * self.scale).round() as f32,
-        )
+        vger::defs::LocalPoint::new((point.x * self.scale) as f32, (point.y * self.scale) as f32)
     }
 
     fn vger_rect(&self, rect: Rect) -> vger::defs::LocalRect {


dzhou121:
It's pretty much identical 

before 

Screenshot 2024-01-06 at 16 44 52

after

Screenshot 2024-01-06 at 16 45 22

dzhou121:
The rounding is pretty important for svgs as well because they will be blurry if placing on subpixel.

Zoxc:
I think I see the reason why this improves thing. We have to specify the center lines of strokes which are in the middle of a pixel, so the rect has to be at +0.5 for it to be pixel perfect and rounding messes that up. We don't see any difference with 200% scaling since the widget gallery has no 1 pixel lines.

Could you take screenshots with a list border changed to `border(0.5)` before and after this PR?

Zoxc:
> The rounding is pretty important for svgs as well because they will be blurry if placing on subpixel.

Rendering of SVG and images have rounding which is not modified here. They should probably rely on views being pixel aligned though. Text seem to round in Y direction only, which doesn't seem right given that it also has subpixel bins.

dzhou121:
border(0.5) does make a difference.

before PR

Screenshot 2024-01-06 at 19 27 04

after PR

Screenshot 2024-01-06 at 19 28 20

dzhou121:
rounding for text is a special one. It rounds based on 0.25 intervals (`SubpixelBin`). And when Swash does rasterising, it offsets it based on the `SubpixelBin`.

dzhou121:
I see svg isn't affected by this

PR #260, opened 2024-01-07
Expose custom theme functionality

pieterdd:
I noticed Floem had functionality in place to support theming. It wasn't exposed to the end user yet, so I did.

This commit also reworks the `themes` example to utilize it, which unfortunately means its blue example theme is gone for now. If a formal mechanism to switch themes while the window is open is implemented, the blue theme can be reintroduced.


Zoxc:
Stylesheets is the current theming mechanism in Floem and the `themes` example shows how to define themes and switch between them. I suggest leaving that alone.

The API in this PR also doesn't make it clear how to opt-out of the default stylesheet.

pieterdd:
> Stylesheets is the current theming mechanism in Floem and the `themes` example shows how to define themes and switch between them. I suggest leaving that alone.

The stylesheet mechanism is preserved after the change, no? The green theme in the example overrides stylesheets (style classes) defined by the default widgets. The main difference is that instead of applying the style rules to a top-level container, they're applied via `WindowConfig`:

```rust
fn main() {
    Application::new()
        .window(
            |_| app_view(),
            Some(
                WindowConfig::default()
                    .title("Custom Theme Example")
                    .theme(compile_custom_theme()),
            ),
        )
        .run();
}
```

The switching functionality is indeed lost because switching themes at runtime via `WindowConfig` is trickier than what Floem does now. This could be an argument to keep applying themes on the top-level container, although it would still require setting `themed` to false when initializing the window - and that feels like I'm configuring my theme in two places.

> The API in this PR also doesn't make it clear how to opt-out of the default stylesheet.

`WindowTheme::default().style(Style::new())` would replace the default theme (I assume that's what you mean with stylesheet in this context) with no style at all.

I'm not confident that what I have is the best possible API, so I'm open to alternatives. I landed on the current version because I saw that `WindowConfig` already exposed a `themed` property that's used to inject the default theme on window level - as opposed to top-level container, which is where the themes example previously applied its rules. I assumed that replacing `themed: bool` with `theme: WindowTheme` was a logical extension of that, allowing for applications to use the default theme (`WindowConfig::default` does) or a completely custom one.

pieterdd:
If you feel like `WindowConfig` shouldn't concern itself with theming, another option could be to remove `themed: bool` from it, not have the window itself paint a theme background, and leave it up to the top-level container to apply the theme (including a background). I'm not completely sure that this would provide good developer ergonomics though, as it would require users of the default theme to add `.apply(default_theme())` to their window's top-level container every time.

jrmoulton:
I'm currently ok with the current method for adding theming. I don't think it necessarily is an improvement to pass a theme style into the window config. I've been applying my theme style with classes to my root view and that works well and to me seems clear/intuitive/flexible.

pieterdd:
Thank you for your feedback. I'll close this PR and clarify in the documentation that to use/make a custom theme, users should apply the theme's style to the root view and if desired disable the default theme in WindowConfig.

PR #261, opened 2024-01-07
Clarify workflow for theming Floem apps

pieterdd:
I'm a bit on the fence about whether the second commit is worth a breaking change and maybe too verbose, but I do think it's clearer. I can still drop it if `themed` is clear enough.


jrmoulton:
I'm in favor of the added clarity. I was confused by this for a while as well. 

Maybe apply_default_theme(bool)?

pieterdd:
`apply_default_theme` works for me! I'll update the code.

PR #280, opened 2024-01-20
Add flight booker example from 7gui

sudormrfbin:
This is an example that emulates an application that books flights, as described in [task 3][task3] of [7gui tasks][7gui].

> The focus of Flight Booker lies on modelling constraints between widgets on the one hand and modelling constraints within a widget on the other hand. Such constraints are very common in everyday interactions with GUI applications. A good solution for Flight Booker will make the constraints clear, succinct and explicit in the source code and not hidden behind a lot of scaffolding.

[task3]: https://eugenkiss.github.io/7guis/tasks/#flight
[7gui]: https://eugenkiss.github.io/7guis/

### DIFF
--- a/examples/flight_booker/Cargo.toml
+++ b/examples/flight_booker/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "flight_booker"
+edition = "2021"
+license.workspace = true
+version.workspace = true
+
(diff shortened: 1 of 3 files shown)


dzhou121:
Thanks!

PR #283, opened 2024-01-21
Introduce on_update callbacks

pieterdd:
This is related to the [recent discussion](https://github.com/lapce/floem/discussions/273) about how widgets should report variable changes. Although `checkbox(signal).on_change(..)` seemed to be the general preference, we hit some roadblocks while figuring out the practicalities.

The system implemented by this PR is `checkbox(signal, |value| { .. })`. If we decide that being able to move forward is more important than implementing the ideal API, we could merge this PR.

This PR does not yet incorporate a request to expose interaction information (e.g. was the update triggered by mouse/keyboard? where did the user click?) in the `on_update` callback. Depending on the opinion of the maintainers, I could still forward the originating `Event` in `on_update` callbacks.

PR #286, opened 2024-01-22
Add `on_update()` method on Checkbox

sudormrfbin:
This PR tackles the API design proposed in #273, specifically having a separate `on_update()` method:

```rust
let (is_checked, set_is_checked) = create_signal(true);

checkbox(is_checked)
	.on_update(move |checked| set_is_checked.set(checked));
```

### Implementation

This is achieved by creating a new `Checkbox` struct that wraps another `View` and simply provides the `on_update()` convenience method. The struct itself implements the `View` trait using a new macro `delegate_view!()` that simply delegates the appropriate methods to the actual inner view (the macro definition was copied almost verbatim from the `View` impl on `Container`).

### Considerations

The amount of code required to use this approach (from the maintainer perspective) is definitely more than simply accepting a callback in the view functions (as showcased in #283), but aside from the nicer end-user API this PR provides, having dedicated View structs allow us to add more chainable methods in the future. 

### Future Work

- Port radio buttons
- Use the new `delegate_view!()` macro to simplify `Container`.

### DIFF
--- a/examples/widget-gallery/src/checkbox.rs
+++ b/examples/widget-gallery/src/checkbox.rs
@@ -24,18 +24,15 @@ pub fn checkbox_view() -> impl View {
                     .disabled(|| true)
             }),
             form_item("Labelled Checkbox:".to_string(), width, move || {
-                labeled_checkbox(is_checked, || "Check me!").on_click_stop(move |_| {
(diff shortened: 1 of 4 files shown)


sudormrfbin:
Closing in favour of #290.

PR #287, opened 2024-01-22
add value_container for provide on_update to checkbox

dzhou121:


src/widgets/checkbox.rs
```diff
@@ -18,17 +18,41 @@ fn checkbox_svg(checked: ReadSignal<bool>) -> impl View {
...
+    create_effect(move |_| {
+        let checked = checked();
+        checked_signal.set(checked);
+    });
+
+    let update_checked_signal = create_rw_signal(default_checked);
+    create_effect(move |_| {
+        let checked = update_checked_signal.get();
+        checked_signal.set(checked);
+    });
```
pieterdd, line 42: I wonder if there's a way to simplify this part. Since any composited widget that returns a ValueContainer will follow the pattern set up here, any simplification will pay off for Floem devs as well as users who make their own components.

jrmoulton:
This looks good to me

src/widgets/checkbox.rs
dzhou121: This part is probably quite different for different views I think.

pieterdd: When I get back from work tonight, I'd like to try rewriting this widget so it doesn't require additional signals and effects. I might arrive at the same conclusion, but it would be satisfying to achieve a simpler result.

pieterdd: It looks like I reduced the amount of additional signals from two to one and preserved the original ReadSignal parameter without breaking anything:

https://github.com/lapce/floem/assets/1407980/b703ae77-b328-4d56-8b0d-7b9944fda5f6

```rust

/// Renders a checkbox using the provided checked signal.
pub fn labeled_checkbox<S: Display + 'static>(
    checked: ReadSignal<bool>,
    label: impl Fn() -> S + 'static,
) -> ValueContainer<bool> {
    let inner_signal = create_rw_signal(checked.get_untracked());
    create_effect(move |_| {
        if checked.get() != inner_signal.get_untracked() {
            inner_signal.set(checked.get());
        }
    });

    value_container(
        h_stack((checkbox_svg(inner_signal.read_only()), views::label(label)))
            .class(LabeledCheckboxClass)
            .style(|s| s.items_center().justify_center()),
        move || inner_signal.get(),
    )
    .keyboard_navigatable()
    .on_click_stop(move |_| {
        inner_signal.set(!checked.get());
    })
}
```

If this aligns with your intentions, I might be able to further simplify the implementation - specifically the part about the inner signal that propagates changes from the input events to the `on_update`. It might be possible to create a helper function for this to minimize repetitive code.

src/widgets/checkbox.rs
```diff
@@ -18,17 +18,41 @@ fn checkbox_svg(checked: ReadSignal<bool>) -> impl View {
 
 /// Renders a checkbox the provided checked signal.
 /// Can be combined with a label and a stack with a click event (as in `examples/widget-gallery`).
-pub fn checkbox(checked: ReadSignal<bool>) -> impl View {
-    checkbox_svg(checked).keyboard_navigatable()
+pub fn checkbox(checked: impl Fn() -> bool + 'static) -> ValueContainer<bool> {
+    labeled_checkbox(checked, || "")
```
pieterdd, line 22: This is not quite the same as before. Here's the new version:

image

I'd consider keeping the separate implementation.

src/widgets/checkbox.rs
pieterdd: Here's a new iteration that builds on your work: https://github.com/lapce/floem/pull/289

```rust
/// Renders a checkbox using the provided checked signal.
pub fn labeled_checkbox<S: Display + 'static>(
    checked: ReadSignal<bool>,
    label: impl Fn() -> S + 'static,
) -> ValueContainer<bool> {
    let (inner_signal, set_inner_signal) = create_inner_signal(checked);

    value_container(
        h_stack((checkbox_svg(inner_signal), views::label(label)))
            .class(LabeledCheckboxClass)
            .style(|s| s.items_center().justify_center())
            .keyboard_navigatable()
            .on_click_stop(move |_| {
                set_inner_signal.set(!checked.get());
            }),
        inner_signal,
    )
}
```

dzhou121: > Here's a new iteration that builds on your work: #289
> 
> ```rust
> /// Renders a checkbox using the provided checked signal.
> pub fn labeled_checkbox<S: Display + 'static>(
>     checked: ReadSignal<bool>,
>     label: impl Fn() -> S + 'static,
> ) -> ValueContainer<bool> {
>     let (inner_signal, set_inner_signal) = create_inner_signal(checked);
> 
>     value_container(
>         h_stack((checkbox_svg(inner_signal), views::label(label)))
>             .class(LabeledCheckboxClass)
>             .style(|s| s.items_center().justify_center())
>             .keyboard_navigatable()
>             .on_click_stop(move |_| {
>                 set_inner_signal.set(!checked.get());
>             }),
>         inner_signal,
>     )
> }
> ```

If `on_update` updates `checked`, it will update `inner_singal`, and `inner_singal` will trigger `on_update` which ends up a infinite loop?

pieterdd: I implemented a stop condition for that purpose:

```rust
pub fn create_inner_signal<T>(outer_signal: ReadSignal<T>) -> (ReadSignal<T>, WriteSignal<T>)
where
    T: Clone + PartialEq + 'static,
{
    let (inner_signal, set_inner_signal) = create_signal(outer_signal.get_untracked());
    create_effect(move |_| {
        if outer_signal.get() != inner_signal.get_untracked() {
            set_inner_signal.set(outer_signal.get());
        }
    });

    (inner_signal, set_inner_signal)
}
```

For me, the widget gallery example of the checkbox works with this rule in place. Are you able to reproduce the infinite loop on my branch?

dzhou121: Yeah this check prevents an infinite loop. 

I would prefer checked to be a closure though. So that for a checkbox I'm only interested in the value update I can do
```
checkbox(|| false)
```
without the need to create a signal

pieterdd: I think I can make that work, but probably not today.

pieterdd: Just to make sure I understand the `checkbox(|| false)` use case.

My working assumption is that event handling would work like this:
- Checkbox is initialized to `false`
- User clicks checkbox
- Checkbox triggers on_update event with value `true`
- Creator of the checkbox updates the signal value to `true`
- Checkbox is notified of the signal value change through reactivity
- Checkbox reflects the `true` state

If you initialize your checkbox with `checkbox(|| false)`, won't it perpetually be stuck in an off state?

dzhou121: That's why there were 2 signals internally. One is "input" and the other is "output".

pieterdd: That part I understand, but in what situation would someone need a closure instead of a ReadSignal? As far as I'm aware, `checkbox(|| false)` is a checkbox that can never be checked. So I'm not sure I understand the use case.

dzhou121: > That part I understand, but in what situation would someone need a closure instead of a ReadSignal? As far as I'm aware, `checkbox(|| false)` is a checkbox that can never be checked. So I'm not sure I understand the use case.

When you click it, it will trigger `on_update`

jrmoulton: > That part I understand, but in what situation would someone need a closure instead of a ReadSignal? As far as I'm aware, `checkbox(|| false)` is a checkbox that can never be checked. So I'm not sure I understand the use case.

That closure will only be run once since it doesn't contain a signal. So it will only set the initial state. If it contained a signal then it would set it every time that the signal changed

sudormrfbin: > That part I understand, but in what situation would someone need a closure instead of a ReadSignal?

One use case that I can think of is for some kind of derived signal -- a simple closure that computes a value using other "real" signals:

```rust
let count = create_rw_signal(0);

checkbox(|| count.get() == 42)
    .on_update(|checked| count.set(42));
```

jrmoulton: > One use case that I can think of is for some kind of derived signal -- a simple closure that computes a value using other "real" signals:
> 
> ```rust
> let count = create_rw_signal(0);
> 
> checkbox(|| count.get() == 42)
>     .on_update(|checked| count.set(42));
> ```


I run into this all the time

pieterdd: Thanks everyone for explaining. Now that I understand the purpose, I made the necessary modifications to support `checkbox(|| false)` at https://github.com/lapce/floem/pull/290 and created a generic function `create_value_container_signals` with documentation to make it reusable. I hope it's useful.

dzhou121:
close in favour of #290

PR #290, opened 2024-01-24
Iteration 3 of 'Add value_container for provide on_update to checkbox'

pieterdd:
Building on dzhou121's previous work, this is another iteration of the ValueContainer that enables `on_update` callbacks. It also returns the unlabeled checkbox to its original implementation to re-align the layout:

image

The primary difference with my second iteration is that `checkbox(|| false)` is supported.


dzhou121:
LGTM. Thanks!

PR #292, opened 2024-01-25
Refactor radio button to use value container

pieterdd:


src/widgets/radio_button.rs
```diff
@@ -24,27 +27,47 @@ where
...
+pub fn radio_button<T>(
+    represented_value: T,
+    actual_value: impl Fn() -> T + 'static,
+) -> ValueContainer<T>
 where
-    T: Eq + PartialEq + Clone + 'static,
+    T: Eq + PartialEq + Copy + 'static,
 {
-    radio_button_svg(represented_value, actual_value).keyboard_navigatable()
+    let (inbound_signal, outbound_signal) = create_value_container_signals(actual_value);
```
dzhou121, line 37: I think we should make the `T` in `create_value_container_signals` `Clone` rather than `Copy`.

pieterdd: Done. I kept the `Copy` of the `T` in the radio button functions though, since I have to move a copy of `represented_value` into a closure there.

dzhou121: You can create a clone for it as well. `Copy` is just so strict here.

e.g.
```
let local_represented_value = represented_value.clone();
value_container(
        radio_button_svg(represented_value, inbound_signal.read_only())
            .keyboard_navigatable()
            .on_click_stop(move |_| {
                outbound_signal.set(local_represented_value.clone());
            }),
        move || outbound_signal.get(),
    )
```

pieterdd: Hmm, I didn't realize Rust treated a `let` differently than a function parameter. Thanks for the tip, I think I fixed it.

PR #301, opened 2024-02-02
change the View Widget trait to IntoView and View

dzhou121:
This PR changes the `View` and `Widget` trait to `IntoView` and `View` respectively. It also changes the `IntoView` trait behaviour to not accessing `ViewData` but we can now impl `IntoView` for `()` `&str` `Vec` etc. 

It also unifies the Views and Widgets into Views. I find `views/list` `widgets/list` rather confusing.


Zoxc:
This also removes the `widgets` and `views` module distinction. The difference there is that the `widgets` module is themed while `views` are not. That's unrelated to the `View` and `Widget` traits as they're all views.

dzhou121:
Do we really need the distinction though? I don’t feel it’s a problem if everything is themed (actually you’d need a default theme for everything, e.g. a basic label still needs the default font color and background otherwise how do you know which color to draw them).

dzhou121:
@Zoxc do you think that removing widget data access from IntoView would be a problem? E.g. something possible before isn’t possible after the change.

jrmoulton:
I'm also fine with removing the views/widget distinction. I don't think that it is problematic to apply classes to the items in the view module. I always remove the default theme but use the widgets module items in order to apply class styles

Zoxc:
> Do we really need the distinction though?

We'd want a way to remove theming if we don't have that distinction.  I think something like style clearing `class` could work.

jrmoulton:
> We'd want a way to remove theming if we don't have that distinction.

More than just setting `apply_default_theme` to false in the window config?

Zoxc:
The removal of `ViewData` from the `View` trait seems to be problematic for a few reasons:
- End users now have to build the view before getting access to the view decorator methods.
- End users now have to care about what is a `IntoView` and what is a `View`, when previously everything was a `View`. Sometimes they would need to build the view and other times not.
- This seems to effectively merge `View` and `Widget` back together and adding a new `IntoView` trait more like a literal `Into<V>` for `V: View` which erases the separation between views and widgets. Part of the motivation here is to hide the `Widget` trait from end users making it easier to potentially swap over to Masonry's `Widget` trait.

It would make sense to have `IntoView` in addition to `View` and `Widget` traits as that would let you convert strings into views as you do in this PR. That's ortogonal to the `View` / `Widget` split.

Zoxc:
> More than just setting `apply_default_theme` to false in the window config?

You may want to use the default theme, but style a specific widget in particular way. It's quite awkward to complete clear the built-in style currently.

dzhou121:
> * End users now have to build the view before getting access to the view decorator methods.

Yeah this is the annoying part.

> * End users now have to care about what is a `IntoView` and what is a `View`, when previously everything was a `View`. Sometimes they would need to build the view and other times not.

We could let everything be an `IntoView`. Any examples that when a user doesn't need to build the view? 

> Part of the motivation here is to hide the Widget trait from end users making it easier to potentially swap over to Masonry's Widget trait.

We can change everything to `IntoView` and hide the `View` trait? Also, would it still be relevant for `ViewData` if we swap over to Masonry?

Zoxc:
> We could let everything be an `IntoView`

Not sure what you mean here by `everything`.

I'm also not sure having an `IntoView` which works for strings, etc. is very compelling. You'd often need to apply styles to them then there's no longer an ergonomic win. It could still work for some cases though. It doesn't seem to be particular harmful other than an odder additional way to do things.

> Also, would it still be relevant for `ViewData` if we swap over to Masonry?

We'd still need a way to access at least `Id`, which follows the same ownership flow as `ViewData`.

I also think we may want typed `Id`s, like having `Id<Scroll>` that has a `scroll_to` method. That may be relevant here.

dzhou121:
I've now brought `ViewData` back to `IntoView`, but instead let user to impl both `IntoView` and `View`, I've impl `IntoView` for `View`

Zoxc:
I think it may be a good idea to split up this PR. It's a bit hard to follow with semantic changes and renames all in one.

Merging the `widgets` module into `views` could be separated out.

Adding a `IntoView` with impls for tuples and string which converts to types with the current `View` trait can be a separate PR too.

dzhou121:
I really don't want to do them again :) 

Basically the only "real" change is removing `any` from `IntoView` (previously `View`), and `build()` changed from returning `Box<dyn Widget>` to `impl Widget`.

Zoxc:
I do see two major problems with this approach. I'm using the current `View` and `Widget` trait names in this comment.

- The removal of `AnyView` means end users now have to convert to the `Widget` layer to erase types. That's quite confusing as currently end users can operate entirely in the `View` layer and they don't need to know about the `Widget` trait unless they want to implement their own widgets.

- `View` has an impl for any `Widget` (`impl<V: View> IntoView for V` in the PR). That prevents any other implementations of `View` due to trait coherence, and such effectively makes the `View` and `Widget` distinction meaningless. The rules of the PR are effectively *all widgets are views* and *you cannot make things that are only views*.

dzhou121:
> * `View` has an impl for any `Widget` (`impl<V: View> IntoView for V` in the PR). That prevents any other implementations of `View` due to trait coherence, and such effectively makes the `View` and `Widget` distinction meaningless. The rules of the PR are effectively _all widgets are views_ and _you cannot make things that are only views_.

I’ll also use `View` and `Widget` trait names here. My understanding is that View will be what the end user mainly interact with, and that will be “fixed”, then we will have different Widget implementations e.g. floem ones or potentially Masonry ones. And this PR doesn’t prevent you only impl View trait but not impl Widget trait. It only makes it that Widgets have to be Views, but I don’t see that a problem, because we won’t swap the View trait but only potentially the Widget trait?

Zoxc:
I think I misremembered the coherence rules. You just can't implement both `View` and `Widget` for the same type, so that won't be problematic.

One thing of note though is that widgets can't be views if we use Masonry widgets, as they won't have `ViewState` like the current design. However it should be fine to use that `impl` until we switch over to avoid the extra `View` `impl`s.

src/view.rs
```diff
@@ -94,94 +94,97 @@ use crate::{
...
 
 pub use crate::view_data::ViewData;
 
-pub type AnyWidget = Box<dyn Widget>;
+pub type AnyView = Box<dyn View>;
 
-pub trait View: Sized {
-    fn view_data(&self) -> &ViewData;
-    fn view_data_mut(&mut self) -> &mut ViewData;
+pub trait IntoAnyView {
```
Zoxc, line 105: What's the purpose of this trait?

src/view.rs
```diff
@@ -94,94 +94,97 @@ use crate::{
...
-            view: Box::new(self),
-        }
-    }
-}
+    fn data_mut(&mut self) -> &mut ViewData;
 
-/// This is a type which can hold any view.
-pub struct AnyView {
-    view: Box<dyn DynView>,
+    fn into_view(self) -> impl View;
```
Zoxc, line 120: You should keep the note on scopes as this seems to be a rename of the current `View::build`.

src/view.rs
```diff
@@ -94,94 +94,97 @@ use crate::{
...
+pub type AnyView = Box<dyn View>;
 
-pub trait View: Sized {
-    fn view_data(&self) -> &ViewData;
-    fn view_data_mut(&mut self) -> &mut ViewData;
+pub trait IntoAnyView {
+    fn any(self) -> AnyView;
+}
 
-    fn id(&self) -> Id {
```
Zoxc, line 108: This method is missing in `IntoView`. That matter for types which are `IntoView`, but doesn't implement `View`.

src/view.rs
```diff
@@ -94,94 +94,97 @@ use crate::{
...
+/// ToView is for converting some types to Floem Views
+pub trait ToView {
+    fn to_view(self) -> impl View;
 }
 
-impl<V: View> DynView for V {
-    fn view_data(&self) -> &ViewData {
-        self.view_data()
+impl ToView for () {
+    fn to_view(self) -> impl View {
```
Zoxc, line 143: This should then return `impl IntoView` as end users operate on `IntoView` and we can't assume that all `IntoView` are also `View`. This would effectively be `IntoIntoView` with your rename. That does suggest that the `View` to `IntoView` rename is a bad idea.

src/views/list.rs
```diff
@@ -62,18 +65,21 @@ where
...
-            data: ViewData::new(Id::next()),
-            selection,
-            index,
-            child: Box::new(v),
-        }
-        .on_click_stop(move |_| {
-            if selection.get_untracked() != Some(index) {
-                selection.set(Some(index));
-                id.update_state(ListUpdate::Accept);
+        container(
```
Zoxc, line 68: This container should be removed. You can just apply the class directly to `Item`.

src/views/virtual_list.rs
```diff
@@ -120,18 +121,25 @@ where
...
-            Item {
-                data: ViewData::new(Id::next()),
-                selection,
-                index,
-                child: Box::new(view_fn(e)),
-            }
-            .on_click_stop(move |_| {
-                if selection.get_untracked() != Some(index) {
-                    selection.set(Some(index))
+            container(
```
Zoxc, line 124: This container should be removed too.

src/views/virtual_list.rs
```diff
@@ -120,18 +121,25 @@ where
...
+                .on_click_stop(move |_| {
+                    if selection.get_untracked() != Some(index) {
+                        selection.set(Some(index))
+                    }
+                })
+                .style(|s| s.width_full()),
+            )
+            .class(ListItemClass)
+            .style(move |s| match direction {
+                VirtualDirection::Horizontal => s.flex_row(),
```
Zoxc, line 140: Then this would become `height_full()` and `width_full()` instead of `flex_row()` and `flex_col()` on `Item` instead.

src/window_handle.rs
```diff
@@ -78,7 +80,7 @@ pub(crate) struct WindowHandle {
 impl WindowHandle {
     pub(crate) fn new(
         window: floem_winit::window::Window,
-        view_fn: impl FnOnce(floem_winit::window::WindowId) -> AnyView + 'static,
+        view_fn: impl FnOnce(floem_winit::window::WindowId) -> AnyView,
```
Zoxc, line 83: Any reason this is changed from `View` to `Widget`?

PR #302, opened 2024-02-03
Fix up dropdown

jrmoulton:


jrmoulton:
@Zoxc 

I've added an on_accept to list.

Issue #300, opened 2024-01-29
proposal: add a way to get application or window handle


Vexcited:
Hi!

I'm trying to see if I can use this with <https://github.com/tauri-apps/window-vibrancy> using their [winit](https://github.com/tauri-apps/window-vibrancy/blob/dev/examples/winit.rs) example.

To use `apply_mica`, I need the window handle (from <https://github.com/lapce/floem/blob/88d6b600f3ef58713682464b95e2c6300821c55f/src/app_handle.rs#L293>) to pass it in the function (<https://github.com/tauri-apps/window-vibrancy/blob/fd1586562868a4b53816a77a25c17c0e3fa354a0/examples/winit.rs#L29>).

Is there any way to make it possible somehow ? Or is there already any API in the library that allows me to get the raw window ?

Thanks you!

lewis-carson:
Hi,

Do you need to use this with window vibrancy or would implementing apply_vibrancy and apply_blur work as well?

Vexcited:
Hi, so to use apply_vibrancy and apply_blur I need to pass the window to the function.

Even if we want to implement those functions ourselves, we would still need to have access to the window ?

Except if those functions are implemented directly in the library... but not sure if that will be done because it's not something everyone needs I guess (?)

lewis-carson:
Looking into this, it looks like window handles are handled by a [fork of winit](https://github.com/lapce/winit), meaning Tauri's window vibrancy would probably need to be re-implemented within Floem's codebase to use Lapce's fork. It doesn't look like a massive amount of effort, could probably be done within a few hours.

I think it's the sort of thing which has enough appetite to justify adding it to Floem. I'll have a look and see if it's something I can do this weekend.

lewis-carson:
I've started the basis for what will turn into window vibrancy. I've forked the [original window vibrancy](https://github.com/tauri-apps/window-vibrancy) and changed it to work with Floem's version of winit. I've added this as a dependency to Floem and started creating a method to set vibrancy within a view. 

I'll open a pull request to track my progress.

lewis-carson:
@Vexcited I have a fork which delivers this functionality but I'm waiting on a pull to be reviewed.

jrmoulton:
You can now get access to the window handle through the `PaintCx` when using Floem from the main branch

PR #303, opened 2024-02-04
Started implementing window vibrancy

lewis-carson:
Issue #300 proposes adding support for changing window vibrancy/adding acrylic effect. I think this can be done with minimal effort and changes to the existing codebase. While this isn't ready to be merged by far, I think it's a good start and can be worked into a functional update with a few hours more work.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ floem_reactive = { path = "reactive", version = "0.1.0" }
 floem-winit = { version = "0.29.4", features = ["rwh_05"] }
 image = { version = "0.24", features = ["jpeg", "png"] }
 copypasta = { version = "0.10.0", default-features = false, features = ["wayland", "x11"] }
+floem-window-vibrancy = "0.4.3"
 
 [features]
 serde = ["floem-winit/serde"]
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -44,6 +44,7 @@ floem_reactive = { path = "reactive", version = "0.1.0" }
 floem-winit = { version = "0.29.4", features = ["rwh_05"] }
 image = { version = "0.24", features = ["jpeg", "png"] }
 copypasta = { version = "0.10.0", default-features = false, features = ["wayland", "x11"] }
+floem-window-vibrancy = "0.4.3"
```
panekj, line 47: Why does it have to be separate crate?

panekj: Also it would've been *nice* to ask us first about publishing crate with `floem` name

lewis-carson:
> Looking at the window vibrancy crate it looked like in order to apply the effect the code just needs access to the raw window handle.

I've deleted the fork; sorry for being presumptuous. I've updated the pull request to allow the user to obtain the raw window handle for their own usage. It should be functional but I've assumed that the window handle doesn't change over the lifespan of a window.

jrmoulton:
This looks fine to me. I don't love the name of the function in decorator. `raw_window_handle` isn't super descriptive but I can't think of anything better. 

@dzhou121

dzhou121:
This approach feels a bit hacky for me. Ideally it should be one option in WindowConfig, and add the vibrancy support to our winit fork.

panekj:
I don't think Floem should expose `WindowHandle` at all

PR #304, opened 2024-02-06
Stack event handlers

jrmoulton:


src/views/decorator.rs
```diff
@@ -96,41 +96,47 @@ pub trait Decorators: View + Sized {
     ///
     /// NOTE: View should have `.keyboard_navigable()` in order to receive keyboard events
     fn on_key_down(
```
Zoxc, line 98: `on_key_down` and `on_key_up` are already stackable.

jrmoulton: I removed the handlers from view data and made them just work like all of the other event handlers. Is there a reason to leave it this way?

Zoxc: Yes. They have less startup cost as they directly mutate `ViewData` and don't need to send an update message.

jrmoulton: Should any other events be implemented with handlers in ViewData?

src/context.rs
```diff
@@ -903,21 +905,19 @@ impl<'a> EventCx<'a> {
             _ => (),
         }
 
-        for handler in &view.view_data().event_handlers {
```
Zoxc, line 906: You should keep this handling. These are supposed to be unconditional and trigger once on any unhandled event.

src/views/decorator.rs
Zoxc: I'd move as many as possible to use `ViewData.event_handlers`. Just have to be careful to not change behavior. I'd probably add a `ViewData::add_event_handler` method which boxes and pushes the handler for convenience.

Keep that in a separate PR if you give it a shot.

PR #305, opened 2024-02-06
Add stacking classes

jrmoulton:


Zoxc:
Probably should rename `update_class` to `add_class`.

PR #308, opened 2024-02-08
Use the WGPU_BACKEND env var if applicable

tamewild:
This allows the specification of the wgpu backend through the `WGPU_BACKEND` environmental variable if present. I believe this is the default behavior in iced as well.

The default backend for my system, vulkan, is really bad and pretty unusable so I need to set it to opengl. I don't think you can specify the backend in the current version of floem so I hope this is an acceptable solution for now at least


tamewild:
seems i forgot to fmt

dzhou121:
Thanks!

PR #311, opened 2024-02-09
Styling::font_size is called using return value of previous Styling::font_size call.

andystopia:
I was trying to build an Obsidian like markdown editor, and I came across this line and making the change provided seemed to allow me to control the font size per line (hardcoded). Please let me know if I've done something wrong. The editor is extremely impressive, and I'm loving the library; I hope a change this trivial isn't burdensome for the devs.

### DIFF
--- a/src/views/editor/mod.rs
+++ b/src/views/editor/mod.rs
@@ -1085,7 +1085,7 @@ impl TextLayoutProvider for EditorTextProv {
 
         let line_content_original = text.line_content(line);
 
-        let font_size = self.style.font_size(self.style.font_size(line));
+        let font_size = self.style.font_size(line);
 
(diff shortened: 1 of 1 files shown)


MinusGix (approved):
Oops, I love making typos, PR looks good to me. Sorry that I didn't see this before, my notifications are a mess.
(Also I'm interested in that you're making an Obsidian-like editor, I've been wanting to see a native version of that for a while. I have a partially-working Katex port for Rust that still needs work and some form of integration with floem but would be useful for full feature parity if you get far enough. I'm potentially willing to work on finishing that up sometime)
I'm not sure that all the code will work properly with different font sizes per line, as it hasn't been tested as much, but it might be fine.

PR #313, opened 2024-02-11
Make extension event handler public

jrmoulton:
Makes it possible to create signals from async tasks

Didn't look like this would be problematic to me

### DIFF
--- a/src/ext_event.rs
+++ b/src/ext_event.rs
@@ -10,7 +10,7 @@ use crate::{
     Application,
 };
 
-pub(crate) static EXT_EVENT_HANDLER: Lazy<ExtEventHandler> = Lazy::new(ExtEventHandler::default);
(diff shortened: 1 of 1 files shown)


Zoxc:
A function which returns it probably makes more sense.

src/ext_event.rs
```diff
@@ -34,6 +34,10 @@ impl ExtEventHandler {
     }
 }
 
+pub fn add_ext_trigger(trigger: Trigger) {
```
dzhou121, line 37: I'm wondering if it's better to let the function to create the `Trigger` and return it?

jrmoulton: Probably about the same to me. But actually now that I'm thinking again I would take the trigger as an argument but call the function register_ext_trigger. 

Either works for me

dzhou121: `register_ext_trigger` does look better

PR #326, opened 2024-02-18
Fixed up file dialog

dominikwilkowski:
As far as I can see most of the methods on `FileDialogOptions` should be cleaned up. Not sure what they are for so I left them alone for now. Happy to remove them in this PR as well.

The main reason for this PR is to add `title`, `allowed_types` and `name`, `multi_selection` support to `FileDialogOptions` and pass that on to `rfd::FileDialog`.
While I was at it I also added an example for the file dialogs as pictured below (mainly also so I can test things are working).
image

### DIFF
--- a/src/action.rs
+++ b/src/action.rs
@@ -117,6 +117,12 @@ pub fn open_file(
         if let Some(path) = options.starting_directory.as_ref() {
             dialog = dialog.set_directory(path);
         }
+        if let Some(title) = options.title.as_ref() {
+            dialog = dialog.set_title(title);
+        }
+        if let Some(filter) = options.filter.as_ref() {
+            dialog = dialog.add_filter(filter.name, filter.extensions);
+        }
         let path = if options.select_directories {
             dialog.pick_folder()
         } else {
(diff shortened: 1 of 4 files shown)


src/action.rs
```diff
@@ -109,32 +109,66 @@ pub fn open_file(
     options: FileDialogOptions,
     file_info_action: impl Fn(Option<FileInfo>) + 'static,
 ) {
-    let send = create_ext_action(Scope::new(), move |path: Option<PathBuf>| {
-        file_info_action(path.map(|path| FileInfo { path, format: None }))
-    });
+    let send = create_ext_action(
+        Scope::new(),
+        move |(path, paths): (Option<PathBuf>, Option<Vec<PathBuf>>)| {
```
dominikwilkowski, line 114: I needed to juggle the two different types you get with `multi_selection` enabled. This was the best I could come up with.

src/action.rs
```diff
@@ -109,32 +109,66 @@ pub fn open_file(
...
+            if paths.is_some() {
+                file_info_action(paths.map(|paths| FileInfo {
+                    path: paths,
+                    format: None,
+                }))
+            } else {
+                file_info_action(path.map(|path| FileInfo {
+                    path: vec![path],
```
dominikwilkowski, line 122: The trade off is we always return a Vec even when you can only select a single thing. I personally think that's easier to handle in user land but happy to hear others to chime in

src/file.rs
```diff
@@ -36,7 +36,7 @@ pub struct FileInfo {
 
 impl FileInfo {
     /// Returns the underlying path.
-    pub fn path(&self) -> &Path {
+    pub fn path(&self) -> &Vec<PathBuf> {
```
dominikwilkowski, line 39: Using `Path` here would break the compiler. Since we already type is as `PathBuf` above I saw no reasons to fight the compiler here.

PR #343, opened 2024-03-02
Add recapture button to inspector

MinusGix:
This PR simply adds a 'Recapture' button to the inspector view:  
image

(Q: should it be 'Refresh'? Should the button be smaller? Should it just be a refresh icon, but then we have to package a small svg?)  

This is useful in scenarios where you want to recapture without interacting with the primary window itself.  
An example is a focus issue in Lapce that I'm debugging, where I expect the focus is ending up somewhere it shouldn't, but if I want to trigger the inspector then I need the focus elsewhere. So I can't find out where the focus ends up! Recapture lets me interact with the primary window, recapture, and see where the focus is at.  
  
Why this is a draft pr is that there's a bug with this code where sometimes when pressing recapture it will produce a blank screen, for reasons I don't know.

### DIFF
--- a/src/inspector.rs
+++ b/src/inspector.rs
@@ -643,7 +643,11 @@ struct CaptureView {
     highlighted: RwSignal<Option<Id>>,
 }
 
-fn capture_view(capture: &Rc<Capture>) -> impl View {
+fn capture_view(
+    window_id: WindowId,
+    capture_s: RwSignal<Option<Rc<Capture>>>,
+    capture: &Rc<Capture>,
+) -> impl View {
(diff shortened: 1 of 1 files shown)


Zoxc:
> Why this is a draft pr is that there's a bug with this code where sometimes when pressing recapture it will produce a blank screen, for reasons I don't know.

Do you mean the `Captured Window` panel will be blank? If so, that's an existing issue.

MinusGix:
> Do you mean the `Captured Window` panel will be blank? If so, that's an existing issue.

That's what I meant, thanks for clarifying!

PR #345, opened 2024-03-04
build(deps): update rfd from 0.11.4 to 0.14.0

panekj:


Cargo.toml
```diff
@@ -59,8 +59,12 @@ image = { version = "0.24", features = ["jpeg", "png"] }
 copypasta = { version = "0.10.0", default-features = false, features = ["wayland", "x11"] }
 
 [features]
-default = ["editor"]
-# TODO: this is only winit and the editor serde, there are other dependencies that still depend on 
+default = ["editor", "rfd-tokio"]
```
dzhou121, line 62: why do we need `rfd-tokio` by default?

panekj: also there has to be one runtime selected for `ashpd` (by default it's `async-std`)

dzhou121: But it doesn't need to be default in Floem?

panekj: There should be some kind of default

dzhou121: The default should be sync API, so no tokio, no async-std

panekj: There is no sync API for dbus

panekj: > The XDG Desktop Portal backend is used when the `gtk3` feature is disabled with [default-features = false](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features), and `xdg-portal` is enabled instead. Either the `tokio` or `async-std` feature must be enabled.

dzhou121: Got it. Sorry didn't see that bit.

panekj: Also, in current state, it already is async API, because only `async-std` was implemented before. Since 0.13.0, `tokio` runtime is provided as well.

panekj: > Got it. Sorry didn't see that bit.

No problem :)

PR #347, opened 2024-03-04
build(deps): update resvg from 0.33.0 to 0.40.0

panekj:


src/views/svg.rs
```diff
@@ -67,7 +67,7 @@ impl Widget for Svg {
     fn update(&mut self, cx: &mut crate::context::UpdateCx, state: Box<dyn std::any::Any>) {
         if let Ok(state) = state.downcast::<String>() {
             let text = &*state;
-            self.svg_tree = Tree::from_str(text, &usvg::Options::default()).ok();
+            self.svg_tree = Tree::from_str(text, &usvg::Options::default(), font_db).ok();
```
panekj, line 70: @dzhou121 `resvg` now requires `&fontdb::Database` as 3rd input for the SVG `Tree`, do you have any advice/tip how to provide that without duplicating database init?

panekj: `FONT_SYSTEM.db` is private unfortunately

dzhou121: It has `.db()`

https://github.com/lapce/cosmic-text/blob/main/src/font/system.rs#L88

panekj: ok, it exists but on main, not in `floem` branch, but also it exists in `floem` branch but only for `no_std`

dzhou121: Yeah looking at the wrong thing. We'll need to expose it then I guess.

panekj: (or syncing with upstream would be nice... )
> This branch is [125 commits behind](https://github.com/lapce/cosmic-text/compare/main...pop-os%3Acosmic-text%3Amain) `pop-os/cosmic-text:main`.

dzhou121: Yeah at some point :)

panekj:
Closing in favour of #406

Issue #341, opened 2024-03-01
Editor issue collection


dominikwilkowski:
Happy to create an issue for each if that is better but didn't want to spam the issues

1. When in `light` mode the gutter paints the gutter number funny
  <img width="912" alt="image" src="https://github.com/lapce/floem/assets/1266923/d57ff7da-593b-4d58-8d75-907e285fbbff">
  (In the editor example)

  ```diff
  - let editor_a = text_editor("Hello World!").styling(SimpleStyling::dark());
  + let editor_a = text_editor("Hello World!").styling(SimpleStyling::light());
  ```
2. The cursor of the editor always blinks even when not in focus
3. The editor seems to ignore `request_focus()`
4. The editor ignores `style` function (should either remove it which is hard with Decorator or support it?)
5. When setting `EditorColor::Caret` color that color is only set when in focus, out of focus editor will have a hard coded caret color
6. The Caret changes width depending if it's on a new line or has text
7. I see no way to actually change anything besides color in the editor like caret width etc
8. Line wrap seems to be off by a few pixel
<img width="912" alt="image" src="https://github.com/lapce/floem/assets/1266923/4acf7ab1-404c-4ab4-8518-13de372cd174">

MinusGix:
> The cursor of the editor always blinks even when not in focus
> The editor seems to ignore request_focus()

Yeah, it isn't handling FocusGained/FocusLost like the original `text_input` does. It should be changed to support that somehow.  
I didn't think of those events when implementing it, so it should be simple to make the default implementation of 'should_blink` use that.


> The editor ignores style function (should either remove it which is hard with Decorator or support it?)

Part of the issue is that we want more complex styling for some applications, like syntax highlighting, and you can't just style an individual piece of text or a line as a view.  
But I think it should be possible to have `SimpleStyling` just default to the (for example) color given to the editorview, since it would merely have a constant color *anyway*. Though I'm not sure of the right way to pass that information to the `Editor` itself.  
(I presume stuff like color, font family information, and background color are what you're referring to here, since it works with style functions like `width_full`)  
  
> `keyboard_navigatable` is being ignored

(TODO: i'm confused)

> The Caret changes width depending if it's on a new line or has text

fun

> I see no way to actually change anything besides color in the editor like caret width etc

I don't know everything people want to customize (admittedly didn't even think of caret width as one of those things), but a lot of it is just "hasn't been hooked up yet". Any other things that should be hooked up to be changeable?  
  
[...]

PR #352, opened 2024-03-06
Move Gutter Layout into Gutter Widget

andystopia:
I was browsing the issues for this repository and I came across #341, which lists, as the first item, that the gutter is drawing the number 1. I have given a shot at modifying the existing gutter, and this is the result that I have come up with.

# Gutter Implementation 
 - handles padding internally instead of having the implementation in `editor_gutter`
 - measures the text internally to increase the width of the gutter in order to keep the amount of padding constant.

# The Gutter Display

| Before (33b12e3): | After |
| --- | --- |
|   image| image | 

# Handling Padding

I realize that handling the padding within the widget is a design call, but I found it simpler during development to have a smaller tree of widgets necessary for the gutter, but I would definitely be willing to put effort into factoring the padding back out if the maintainers would like me to. I really don't mind much either way. You have my apologies if this is overstepping and I will do my best to remedy it immediately.

### DIFF
--- a/src/views/editor/gutter.rs
+++ b/src/views/editor/gutter.rs
@@ -3,6 +3,7 @@ use crate::{
     cosmic_text::{Attrs, AttrsList, TextLayout},
     id::Id,
     peniko::kurbo::Point,
+    style::Style,
     view::{AnyWidget, View, ViewData, Widget},
     Renderer,
 };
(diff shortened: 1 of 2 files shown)


MinusGix (approved):
This looks good to me, thanks!

PR #363, opened 2024-03-09
MenuDisplay enum, fixes #354

### LINKED ISSUE #354: dyn_stack problem with Option<MenuDisplay>
Screenshots that explains the problem: https://imgur.com/a/C1ibQaH

I tracked down the issue to dyn_stack at `src/window_handle.rs` line 1471. dyn_stack takes in proper amount of menu items from `context_menu_items`, but iterates less in `view_fn`.

In `src/views/dyn_stack.rs` when opening the context menu 
```
let items = items.into_iter().collect::<SmallVec<[_; 128]>>();
let hashed_items = items.iter().map(&key_fn).collect::<FxIndexSet<_>>();
// 16 entries and 3 separators ->  items.len() = 19,  hashed_items.len()  = 17
// 16 entries and 4 separators -> items.len() = 20,  hashed_items.len()  = 17
// 16 entries and 5 separators -> items.len() = 21,  hashed_items.len()  = 17 
```
So collecting to `FxIndexSet` seems to remove (amount of separators - 1) from the end of total menu items. 

I tried using default hasher at `src/views/dyn_stack.rs` line 16:  

`pub(crate) type FxIndexSet<T> = indexmap::IndexSet<T, BuildHasherDefault<std::hash::DefaultHasher>>;`

but that didn't help.

aalhitennf:
Menu items are currently `Option<MenuDisplay>` and separators are considered as `None`. `dyn_stack` canno't hash? the Nones properly which causes missing menu items. Switching to enum and giving it something as id fixes this.

### DIFF
--- a/src/window_handle.rs
+++ b/src/window_handle.rs
@@ -1279,30 +1279,34 @@ fn context_menu_view(
     };
 
     #[derive(Clone, PartialEq, Eq, Hash)]
-    struct MenuDisplay {
-        id: Option<u64>,
-        enabled: bool,
-        title: String,
-        children: Option<Vec<Option<MenuDisplay>>>,
+    enum MenuDisplay {
+        Separator(usize),
+        Item {
+            id: Option<u64>,
+            enabled: bool,
+            title: String,
+            children: Option<Vec<MenuDisplay>>,
+        },
     }
 
-    fn format_menu(menu: &Menu) -> Vec<Option<MenuDisplay>> {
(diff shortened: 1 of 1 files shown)


jrmoulton:
diff looks large but the change looks good to me dzhou121

aalhitennf:
> diff looks large but the change looks good to me dzhou121

yep menu items are created with this huge chunk of code inside if statement, now match instead, shifting it by few line/ident or two which makes it look larger than what it actually does

PR #378, opened 2024-03-15
Add method to stack/list to add class by idx

jrmoulton:
Couldn't figure out how to apply it to every view in Decorators because of the view/widget split so I've just added it to Stack and List. 

Very useful when dealing with Grid.


jrmoulton:
@dominikwilkowski   if you're using grid this would be useful

PR #379, opened 2024-03-15
don't change focus to on_click under other on_click

### LINKED ISSUE #295: Views on top of inputs don't fire on_clicks
When I have a view that is positioned absolute over an input field the on_clicks on that view won't fire.
Consider this code:
```rust
use floem::{
    reactive::create_rw_signal,
    style::Position,
    view::View,
    views::{container, label, stack, Decorators},
    widgets::text_input,
};

fn app_view() -> impl View {
    let value = create_rw_signal(String::from(""));
    stack((
        container(text_input(value)),
        label(move || "Label on top")
            .style(|s| {
                s.position(Position::Absolute)
                    .inset_top(8.5)
                    .inset_left(7)
                    .background(floem::peniko::Color::rgba8(255, 0, 0, 150))
            })
            .on_click_cont(|_| {
                println!("Click continue");
            })
            .on_click_stop(|_| {
                println!("Click stop");
            }),
    ))
    .style(|s| {
        s.position(Position::Relative)
            .margin(10)
            .height(30)
            .items_center()
            .justify_center()
    })
}

fn main() {
    floem::launch(app_view);
}
```

This will render something like this: 
image

A label with a red background is positioned on top of an input field.
this is what happens when you try to click on the label:
input

Note there is no print output in `stdout` in the terminal below. Neither for `on_click_cont` nor for `on_click_stop`.

jrmoulton:
@dzhou121 

I've been trying to fix the issue where having an on_click handler on top of another on_click handler will move focus to the bottom one and it will receive the click event instead of the one on top. 

This change fixes the issue but I'm not sure I completely understand it. 

The problem comes because both views will get the pointer down and since the bottom one gets it last it gets the focus for the pointer up. 

Review on this would be good

Fixes #295

### DIFF
--- a/src/window_handle.rs
+++ b/src/window_handle.rs
@@ -318,6 +318,22 @@ impl WindowHandle {
                 }
             }
         }
+
+        if let Event::PointerDown(event) = &event {
+            if cx.app_state.focus.is_none() {
+                if let Some(id) = was_focused {
+                    let layout = cx.app_state.get_layout_rect(id);
+                    if layout.contains(event.pos) {
+                        // if the event is pointer down
+                        // and the focus hasn't been set to anything new
+                        // and the pointer down event is inside the previously focusd view
+                        // we then set the focus back to that view
+                        cx.app_state.focus = Some(id);
+                    }
+                }
+            }
+        }
+
         if was_focused != cx.app_state.focus {
             cx.app_state.focus_changed(was_focused, cx.app_state.focus);
         }
--- a/src/context.rs
+++ b/src/context.rs
@@ -702,6 +702,8 @@ impl<'a> EventCx<'a> {
             return EventPropagation::Stop;
         }
 
+        let mut is_down_and_has_click = false;
+
         match &event {
             Event::PointerDown(event) => {
                 self.app_state.clicking.insert(id);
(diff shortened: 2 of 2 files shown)


jrmoulton:
whoops. pushed to wrong branch

src/window_handle.rs
```diff
@@ -318,7 +321,9 @@ impl WindowHandle {
                 }
             }
         }
-        if was_focused != cx.app_state.focus {
+        if (was_focused != cx.app_state.focus) && !handled {
```
dzhou121, line 324: I feel like this is not needed because the PointerDown event propagation was stopped. Can you please test if it fixes the issue without this check?

jrmoulton: This is the part I don't completely understand what is happening. If I remove that check then anytime that an event returns `Stop` (even a different event handler such as PointerDown) the focus will be switched away.

jrmoulton:
Figured it out. There was another issue. If a view has focus and then there is a pointer down event inside of that view (on a child) that returns EventPropagation::Stop the focus will have been cleared without giving the focus to any other view. 

This checks on pointer down if the event is still inside of the old was_focused and nothing else has gained focus it gives focus back to the view that previously had focus.

Issue #393, opened 2024-03-24
[Editor] ignores `style` function


dominikwilkowski:
The editor ignores `style` function right now.
It should either remove it which is hard with Decorator or support it to the extend that the container gets things border and display etc properties?

Maybe we explicitly say no styling for the editor and you have to use a `container` to wrap it to give it styles?
That only works if the container is always `.width_full()` and `.height_full()`?

jrmoulton:
I’m working on an overhaul of editor styling right now

dominikwilkowski:
fixed by #398

PR #398, opened 2024-03-25
Move some editor styling to floem style props

jrmoulton:
This works and I think works really well. 

I'm marking this as a draft because I still need to add docs and update the examples.

@MinusGix review on this would be good especially for things that might need to be changed to work with Lapce. 
Mostly that is around SimpleStyling which now is only for properties that require line information

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -46,6 +46,7 @@ pub trait StylePropValue: Clone + PartialEq + Debug {
 impl StylePropValue for i32 {}
 impl StylePropValue for bool {}
 impl StylePropValue for f32 {}
+impl StylePropValue for usize {}
 impl StylePropValue for f64 {
     fn interpolate(&self, other: &Self, value: f64) -> Option<Self> {
         Some(*self * (1.0 - value) + *other * value)
@@ -169,7 +170,7 @@ pub trait StyleClass: Default + Copy + 'static {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct StyleClassInfo {
     pub(crate) name: fn() -> &'static str,
 }
@@ -182,7 +183,7 @@ impl StyleClassInfo {
     }
 }
 
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Debug)]
 pub struct StyleClassRef {
     pub key: StyleKey,
 }
@@ -451,7 +452,7 @@ macro_rules! prop_extractor {
                 changed
             }
 
-            #[allow(dead_code)]
+           #[allow(dead_code)]
             $vis fn read(&mut self, cx: &mut $crate::context::StyleCx) -> bool {
                 let mut transition = false;
                 let changed = self.read_explicit(&cx.direct_style(), &cx.indirect_style(), &cx.now(), &mut transition);
@@ -612,7 +613,7 @@ impl StyleKey {
     pub(crate) fn debug_any(&self, value: &dyn Any) -> String {
         match self.info {
             StyleKeyInfo::Selector(..) | StyleKeyInfo::Transition => String::new(),
-            StyleKeyInfo::Class(..) => String::new(),
+            StyleKeyInfo::Class(info) => (info.name)().to_string(),
             StyleKeyInfo::Prop(v) => (v.debug_any)(value),
         }
     }
@@ -1715,8 +1716,8 @@ impl Style {
         self.set(ZIndex, Some(z_index))
     }
 
-    /// Allow the application of a function if the option exists.  
-    /// This is useful for chaining together a bunch of optional style changes.  
+    /// Allow the application of a function if the option exists.
+    /// This is useful for chaining together a bunch of optional style changes.
     /// ```rust
     /// use floem::style::Style;
     /// let maybe_none: Option<i32> = None;
@@ -1734,7 +1735,7 @@ impl Style {
         }
     }
 
-    /// Allow the application of a function if the condition holds.  
+    /// Allow the application of a function if the condition holds.
     /// This is useful for chaining together a bunch of optional style changes.
     /// ```rust
     /// use floem::style::Style;
--- a/src/style.rs
+++ b/src/style.rs
@@ -1716,8 +1716,8 @@ impl Style {
         self.set(ZIndex, Some(z_index))
     }
 
-    /// Allow the application of a function if the option exists.
-    /// This is useful for chaining together a bunch of optional style changes.
+    /// Allow the application of a function if the option exists.  
+    /// This is useful for chaining together a bunch of optional style changes.  
     /// ```rust
     /// use floem::style::Style;
     /// let maybe_none: Option<i32> = None;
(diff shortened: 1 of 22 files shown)


MinusGix (commented):
Overall I like the idea/impl.
I was initially envisioning the floem style being inside `SimpleStyling`, but this is fine and works better in various cases.

Somewhat uncertain about changing `phantom_text` to take `&EditorStyle` instead. I was initially envisoning it as "user can downcast `&dyn Styling` if they have a custom impl to get arbitrary styling information".
That's what I'd have done in Lapce if we didn't already have `config` on our document for phantom text coloration (because lapce's phantom text has things like diagnostics and background color).
But I think that's fine enough.

I think it would be good to have some way of doing a decent default dark/light theme, since that seems no longer used. 
(if we're going with "floem should just have very basic styles", I'm admittedly a bit skeptical of that)

src/views/editor/mod.rs
```diff
@@ -1093,12 +1080,16 @@ impl TextLayoutProvider for Editor {
...
+        let phantom_text = doc.phantom_text(edid, &self.editor_style.get_untracked(), line);
         let line_content = phantom_text.combine_with_text(&line_content);
 
         let family = style.font_family(edid, line);
         let attrs = Attrs::new()
-            .color(style.color(edid, EditorColor::Foreground))
+            .color(
+                self.editor_style
+                    .with(|s| s.text_color())
+                    .unwrap_or(Color::BLACK),
```
MinusGix, line 1091: Just have `text_color` return a `Color` and unwrap to default itself

MinusGix: You could also just get-tracked editor style once at the top to make the function cleaner

src/views/editor/mod.rs
```diff
@@ -1113,6 +1104,8 @@ impl TextLayoutProvider for Editor {
 
             let mut attrs = attrs;
             if let Some(fg) = phantom.fg {
+                //TODO: Phantom text color?
+                // attrs = self.editor_style ;
```
MinusGix, line 1108: Phantom text can have individually different colors in Lapce as its used for inlay hints, information, warnings, and errors.
Could have it default to the phantom color.

src/style.rs
```diff
@@ -1715,8 +1716,8 @@ impl Style {
         self.set(ZIndex, Some(z_index))
     }
 
-    /// Allow the application of a function if the option exists.  
-    /// This is useful for chaining together a bunch of optional style changes.  
+    /// Allow the application of a function if the option exists.
+    /// This is useful for chaining together a bunch of optional style changes.
```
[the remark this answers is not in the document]
jrmoulton: Hmm definitely wasn't me intentionally but I'll make sure it get's changed back

src/views/editor/mod.rs
jrmoulton: TextColor is a floem property and it's an Option<Color>. I'm not really sure why though

> You could also just get-tracked editor style once at the top to make the function cleaner

`with` avoids cloning the style struct

src/views/editor/view.rs
```diff
@@ -892,13 +1004,23 @@ impl Widget for EditorView {
...
+            &screen_lines,
+            &self.editor_view_style,
+        );
         let screen_lines = ed.screen_lines.get_untracked();
-        EditorView::paint_text(cx, &ed, viewport, &screen_lines);
-        EditorView::paint_scroll_bar(cx, &ed, viewport);
+        EditorView::paint_text(cx, &ed, viewport, &screen_lines, &self.editor_view_style);
+
+        // TODO: what is going on here? Why is this necessary
+        EditorView::paint_scroll_bar(cx, viewport, &self.editor_view_style);
```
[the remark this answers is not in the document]
jrmoulton: > (Possibly we could/should just remove the custom scrollbar painting code from floem-editor as Lapce has to write its own function on top for extra git map stuff anyway. Unsure)

I'll look at removing it would be more consistent with the rest of floem if it just used the one from the scroll view.

src/views/editor/mod.rs
MinusGix: > TextColor is a floem property and it's an Option<Color>. I'm not really sure why though

To allow you to know whether you can use a default presumably? (Though probably there's some way to just go "use this as the base style for my view")

Could still just do an `impl EditorStyle { fn ed_text_color() ..`

> `with` avoids cloning the style struct

and `EditorStyle` doesn't have that much in it. Though I don't consider this style too bad.

jrmoulton: > and EditorStyle doesn't have that much in it. Though I don't consider this style too bad.

The style structs are surprisingly large. `EditorStyle` is 536 bytes and they grow quickly with more fields. 

> Though I don't consider this style too bad.

I also don't think the `with` adds too much clutter

> Could still just do an impl EditorStyle { fn ed_text_color() ..

👍

jrmoulton:
I figure out how to revert the whitespace changes so this should clearer to review now.

jrmoulton:
I've fixed several issues that I had and now need to actually go back and add docs and update the examples

src/views/editor/gutter.rs
```diff
@@ -3,25 +3,44 @@ use crate::{
...
+
+prop!(pub LeftOfCenterPadding: f64 {} = 25.);
+prop!(pub RightOfCenterPadding: f64 {} = 30.);
+prop!(pub DimColor: Color {} = Color::DIM_GRAY);
+
+prop_extractor! {
+    GutterStyle {
+        accent_color: TextColor,
+        dim_color: DimColor,
+        left_padding: LeftOfCenterPadding,
```
jrmoulton, line 28: added the ability to customize the gutter padding

src/views/editor/gutter.rs
```diff
@@ -152,14 +184,44 @@ impl Widget for EditorGutterView {
                 let mut text_layout = TextLayout::new();
                 if line == current_line {
                     text_layout.set_text(&text, current_line_attrs_list.clone());
+                    if let Some(current_line_color) = self.gutter_style.current_line_color() {
```
jrmoulton, line 187: Added the ability to have the gutter show the highlighted line. This appears continuous with the editor highlighted line.

jrmoulton:
That was a lot more work than I was expecting when I started. But now everything but line/syntax specific properties in the editor can be styled using floem styles and the editor_style method makes it super easy to use and obvious what is stylable.

MinusGix (approved):
Looks good to me :)

dominikwilkowski:
This fixes #393 no? Or are you expecting to move even more styles over?

jrmoulton:
Oh yeah I just forgot to reference it as fixed. This fixes it

src/views/editor/text.rs
```diff
@@ -359,50 +356,64 @@ pub trait Styling {
     }
 }
 
-pub fn default_light_color(color: EditorColor) -> Color {
+pub fn default_light_theme(mut style: EditorCustomStyle) -> EditorCustomStyle {
```
panekj, line 359: Name for light variant was changed but name of dark variant wasn't. Is there some reason for the difference or why was it changed?

PR #401, opened 2024-03-28
text editor updates

jrmoulton:
I've moved the styling of EditorStyle from TextEditor style to EditorView Style. This makes it so that the editor view can be used without needing another parent wrapper. 

I've added triggers to Editor that are triggered when the editor view is focused/loses focus. 

I swapped out the CaretColor prop to follow the floem default CursorColor prop. 

I added a text_editor_keys function that lets you build a text editor with custom key handling. 

On TextEditor I removed a lot of the unnecessary wrappers that were used to build the child view. It's now much simpler. As part of that I inlined the clipping logic into TextEditor. 

I also added a min size to TextEditor. It is much smaller than anyone would actually want (I think) but it doesn't constrain whoever is using it and it now at least will always appear on the screen at least a bit.

### DIFF
--- a/examples/editor/src/main.rs
+++ b/examples/editor/src/main.rs
@@ -1,12 +1,13 @@
 use floem::{
     keyboard::{Key, ModifiersState, NamedKey},
+    peniko::Color,
     reactive::RwSignal,
     view::View,
     views::{
         editor::{
             command::{Command, CommandExecuted},
             core::{command::EditCommand, editor::EditType, selection::Selection},
-            text::{default_dark_color, SimpleStyling},
+            text::{default_dark_color, RenderWhitespace, SimpleStyling},
         },
         stack, text_editor, Decorators,
     },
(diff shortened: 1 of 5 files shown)


jrmoulton:
cargo fmt doesn't touch whitespace but rust-analyzer format on save keeps removing trailing whitespace

MinusGix:
Strange, I don't have that issue with format on save (using RA) unless that's some extra setting to change formatter.

examples/editor/src/main.rs
```diff
@@ -26,11 +27,18 @@ fn app_view() -> impl View {
...
+                .render_white_space(RenderWhitespace::All)
+        });
     let editor_b = editor_a
         .shared_editor()
         .editor_style(default_dark_color)
-        .editor_style(move |s| s.hide_gutter(hide_gutter_b.get()))
+        .editor_style(move |s| {
+            s.hide_gutter(hide_gutter_b.get())
+                .render_white_space(RenderWhitespace::All)
+                .visible_whitespace(Color::BLACK)
```
MinusGix, line 40: Why render it black? (or is this just to show off the extra styles?)

MinusGix (commented):
Looks good, though should
```rust
if self.editor.es.try_update(|s| s.read(cx)).unwrap() {
            self.editor.floem_style_id.update(|val| *val += 1);
            cx.app_state_mut().request_paint(self.id());
        }
```
that be on `EditorView` instead of on `TextEditor`?

examples/editor/src/main.rs
jrmoulton: Oh whoops. That was just testing. I'll remove that. 

I was doing it though because when there is an empty line I couldn't get any selection to show up. Any idea why that is or what would cause it?

MinusGix: As in the texteditor was completely empty, and you couldn't get a selection to show up?  
Or just an empty line in it? (And by 'empty' do you mean 'with whitespace' or 'literally empty'
I'd assume a screenline problem, because I believe selection uses screen lines to decide whether to draw selection. Or just a problem in the logic for drawing selections.

jrmoulton: When there is an empty line (with a newline)
image
Everything is selected but the empty lines aren't showing

MinusGix: Ah, that's a problem before your PR too.  
I think this is probably because I made visual lines not include the newline, and so they need to do an extra +1 on size for empty lines. I'll open a new issue.

jrmoulton: fixed

MinusGix (approved):

PR #405, opened 2024-03-29
Allow to hook application window events

dajoha:
This can be useful to handle application-level global shortcuts.

In the same time I improved the `keyboard_handler` example, in order to demonstrate the global shortcut capability.

### DIFF
--- a/examples/keyboard_handler/Cargo.toml
+++ b/examples/keyboard_handler/Cargo.toml
@@ -8,3 +8,4 @@ version.workspace = true
 
 [dependencies]
 floem = { path = "../.." }
+floem-winit = { version = "0.29.4", features = ["rwh_05"] }
--- a/examples/keyboard_handler/src/main.rs
+++ b/examples/keyboard_handler/src/main.rs
@@ -1,31 +1,62 @@
 use floem::{
+    Application,
     event::{Event, EventListener},
     keyboard::Key,
+    peniko::Color,
+    quit_app,
+    style::CursorStyle,
     unit::UnitExt,
+    views::{text, Decorators, container},
     view::View,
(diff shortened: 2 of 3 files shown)


dominikwilkowski:
I curious how this is different to what we have now?
We have global shortcuts implemented like this: https://github.com/dominikwilkowski/vault/blob/main/src/main.rs#L256-L290 and it seems to work?

dajoha:
Hello @dominikwilkowski , sorry for late response.

1. The main reason to propose this MR is that for now, for any Linux target, global shortcuts do not work (or you must first click on an empty zone in the UI to get focus, which is unacceptable). The reason is quite simple: the main view (the one given to `Application::new().window(|| {})`) is wrapped internally by floem, in order to handle context menus. Maybe a better solution would be to remove this wrap for Linux targets, but I don't know if it's possible.
2. The 2nd reason is that sometimes, depending on the app you are writing, you may want the keyboard events to be handled at first by the application itself, not by the focused element; for example, for now, focused text editors will catch every keyboard event, and there is no way to quit with <kbd>Ctrl-Q</kbd> if you set this shortcut globally.

To summarize, as developpers we should have full control over this important aspect, but I'm certainly not enough involved to propose the best solution.

Btw as you can see, this MR is outdated, but I made a different version [here](https://github.com/dajoha/floem/commit/a12b209062f3a4b53a5d31ca16440897f930b173) where I moved the hook from `Application` to `WindowHandle`, (which allows to use the computed values for the key modifiers). I will take the time to update it soon with this new version.

dominikwilkowski:
I see. This is linux specific then? Good to know. We should check our implementation to work on linux too then :)

PR #406, opened 2024-03-30
Dependency updates

MinusGix:
Initially this started as dependency updates but has turned into updating wgpu as well, because some of this stuff was causing painful duplication.

image had default features enabled which brought in various image dependencies for things like EXR / DDS / HDR / etc. It seems better to have this as a default of "jpeg", "png" (which was possibly intended by the original PR).
The onle one from that list I was uncertain about skipping was `.ico` and `webp`: https://github.com/image-rs/image/blob/main/Cargo.toml#L70
(Or we could do "None" by default but that might confuse users)

412 -> 399

Other stuff, updating wgpu & resvg/usvg -> 385

Updated softbuffer & copypasta to avoid duplicate x11rb/x11rb-protocol.
-> 392
More random stuff -> 369

Moves `rfd` under the two feature flags `rfd-async-std`/`rfd-tokio`. (I don't see a good way to have a simple `rfd` feature with a default that doesn't make it easy to accidentally bring in both.. I think it just should be a line in the docs.)  
I have set this to off by default with the reasoning that 1) it brings in a lot of deps (369 -> 308, partially due to old zbus dependencies) 2) many programs don't bother with open/save dialogs.  

-----

I'm not sure about having the window be `'static` for wgpu. It has this via an `Arc` — so it won't lose the window handle while it has it — but that means when we drop the surface it should drop the window and then everything's fine?  
Possibly need to test with multiple windows or something to ensure there's no leakage, but I think it should be good.

------

I'll need to open another pr for the floem-vger as it needs updates wgpu

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,8 @@ lapce-xi-rope = { version = "0.3.2", features = ["serde"] }
 strum = "0.21.0"
 strum_macros = "0.21.1"
 once_cell = "1.17.1"
+image = { version = "0.24", default-features = false, features = ["jpeg", "png"] }
+im = "15.1.0"
 
 [dependencies]
 sha2 = "0.10.6"
@@ -40,7 +42,6 @@ kurbo = { version = "0.9.5", features = ["serde"] }
 unicode-segmentation = "1.10.0"
 floem-peniko = "0.1.0"
 crossbeam-channel = "0.5.6"
-im = "15.1.0"
 im-rc = "15.1.0"
 serde = { workspace = true, optional = true }
 lapce-xi-rope = { workspace = true, optional = true }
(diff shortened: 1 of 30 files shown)


MinusGix:
Depends on: https://github.com/lapce/cosmic-text/pull/5
https://github.com/lapce/vger-rs/pull/8

Once I've decided on the `'static` bit (and finished up) then it will probably be a:  
Merge cosmic -> I update vger PR -> merge vger PR -> update this PR -> merge this PR

Cargo.toml
```diff
@@ -20,27 +20,31 @@ license.workspace = true
 [workspace.dependencies]
 serde = "1.0"
 lapce-xi-rope = { version = "0.3.2", features = ["serde"] }
-strum = "0.21.0"
-strum_macros = "0.21.1"
+strum = "0.26.2"
+strum_macros = "0.26.2"
 once_cell = "1.17.1"
+image = { version = "0.24", default-features = false, features = ["jpeg", "png"] }
```
panekj, line 26: I think the defaults should be `"jpeg", "png", "ico", "webp", "gitf"` due to popularity but also being able to disable/enable any of them

PR #408, opened 2024-04-01
treat altgr as Modifier

dzhou121:


dajoha:
Hello @dzhou121 , this PR makes the text editor totally unusable with a french keyboard. Indeed, inserting the following chars is done by using "AltGr": 
```
~#{[|`\^@]}
```

dzhou121:
> Hello @dzhou121 , this PR makes the text editor totally unusable with a french keyboard. Indeed, inserting the following chars can be done only by using "AltGr":
> 
> ```
> ~#{[|`\^@]}
> ```
> 
> And hitting a "AltGr+..." keystroke does not work anymore.
> 
> Maybe this can be fixed by just adding `mods.set(Modifiers::ALTGR, false);` here?
> 
> https://github.com/lapce/floem/blob/b538f39764a38c8d190ff960c60496e318b79394/src/views/editor/view.rs#L1201

You are right. Fixed in main now.

Issue #413, opened 2024-04-08
Tooltips not hiding automatically on some cases


Narref95:
Tooltips are not getting removed when you navigate through the app using keyboard or when view is not visible.

Using widget-gallery example in Label tab there is a tooltip, if you show it and then navigate to another menu, tooltip is still visible.

![image](https://github.com/lapce/floem/assets/17275184/fbe376a7-0ace-469d-845e-53fe32fbf3e5)

This is also happening to me while using Lapce (tooltip of Cntrl + F search box is showing after I click the X button)

PR #414, opened 2024-04-09
Fix tooltip behaviour

Narref95:
fix https://github.com/lapce/floem/issues/413

As keyboard_navigatable only works when element is focused I cant use it to remove the tooltip, so I added the tooltip to the new WindowHandle field instead of overlays map (to be able to remove it directly).

New tooltip behaviour:
Hide tooltip when some key is pressed
Hide tooltip when PointerUp / PointerDown


MinusGix (approved):

Long0x0:
The tooltip still shows when dragging something over:



Could you also fix this?

Narref95:
> The tooltip still shows when dragging something over:
> 
> Could you also fix this?

Nice catch! I didn't though about dragging over the tooltip. I will try to fix that too.

jrmoulton:
My thoughts copy/paste from discord. 

> I don't think there is a strong need to special case tooltips from overlays. 
    To me it looks like a lot of duplicated API just so that you can ensure that there is only one tooltip showing at a time. I think a better solution would be to have the events that deactivate the tooltip be better handled. 

> Part of that would be having a ClickedOutside and (probably) KeyEventOutide events

> I guess specifically in this case what is happening is that the PointerLeave event doesn't fire because the underlying view can be swapped out with keyboard events. Maybe we need to send PointerLeave if that happens

Narref95:
> My thoughts copy/paste from discord.
> 
> > I don't think there is a strong need to special case tooltips from overlays.
> > To me it looks like a lot of duplicated API just so that you can ensure that there is only one tooltip showing at a time. I think a better solution would be to have the events that deactivate the tooltip be better handled.
> 
> > Part of that would be having a ClickedOutside and (probably) KeyEventOutide events
> 
> > I guess specifically in this case what is happening is that the PointerLeave event doesn't fire because the underlying view can be swapped out with keyboard events. Maybe we need to send PointerLeave if that happens

ClickedOutside would not be needed for this case as the tooltip needs to be removed when any PointerDown is performed (inside and outside), this has been fixed adding the handle of this event.
I like the idea of adding KeyEventOutside if this can be used to manage the overlay inside the tooltip view implementation and remove the duplicated code added.
KeyEventOutside should be sent to all widgets of the screen instead of the active one?

Narref95:
Fixed tooltip showing when dragging something over it, also remove tooltip when pointer wheel is triggered.

PR set as draft to implement a better solution that reduces code duplication.

Narref95:
- Added the possibility to listen keyboard events without being focused, this gives the ability to close the tooltip when any key is pressed.
This reduces code duplication instead of duplicating overlay logic for tooltips and adds this functionality to new use cases.

PR #417, opened 2024-04-18
Add new text input widget

presiyan-ivanov:


src/views/editor/text.rs
```diff
@@ -721,7 +721,6 @@ pub struct SimpleStylingBuilder {
     indent_style: Option<IndentStyle>,
     tab_width: Option<usize>,
     atomic_soft_tabs: Option<bool>,
-    wrap: Option<WrapMethod>,
```
presiyan-ivanov, line 724: @MinusGix I removed this since it seems its not used anymore - probably left out from the editor styles migration. Can u confirm its fine to remove it?

MinusGix: Yeah looks fine to remove.

examples/widget-gallery/src/clipboard.rs
```diff
@@ -21,7 +21,9 @@ pub fn clipboard_view() -> impl View {
             }),
             form_item("Copy from input".to_string(), 120.0, move || {
                 h_stack((
-                    text_input(text1).keyboard_navigatable(),
+                    text_input(move || text1.get()).on_update(move |val| {
```
MinusGix, line 24: Should presumably still be `.keyboard_navigatable`

src/views/text_input.rs
```diff
@@ -59,872 +67,545 @@ prop_extractor! {
...
-    // Approx max size of a glyph, given the current font weight & size.
-    glyph_max_size: Size,
-    style: Extracter,
-    font: FontProps,
-    cursor_width: f64, // TODO: make this configurable
-    is_focused: bool,
-    last_cursor_action_on: Instant,
-}
+//TODO add to styles
+static IME_FOREGROUND: Color = Color::BLACK;
```
MinusGix, line 71: Probably want to add this to styles

src/views/text_input.rs
```diff
@@ -59,872 +67,545 @@ prop_extractor! {
...
+    let cursor_line = create_rw_signal(Line::new(Point::ZERO, Point::ZERO));
     let is_focused = create_rw_signal(false);
+    let hide_cursor = editor.with_untracked(|e| e.cursor_info.hidden);
+    let skip_content_sync = create_rw_signal(true);
+    let pointer_focus = create_rw_signal(false);
+
+    // Syncs with external changes to the text. This should *only* happen if the closure
+    // re-runs due to an external update.
+    // For example the closure is subscribed to a signal and the signal is updated by another view.
+    // NOTE: The updates are dispatched in the same order they occur, so this is safe to do
```
MinusGix, line 91: I don't think the comment past the first line is really needed, that being what signals do & are for.

src/views/text_input.rs
```diff
@@ -947,426 +627,221 @@ impl Widget for TextInput {
...
-                .width(taffy_node_width)
-                .height(self.height)
-                .to_taffy_style();
-            let _ = cx.app_state_mut().taffy.set_style(text_node, style);
+                let style = Style::new().height(height).to_taffy_style();
+                cx.set_style(text_node, style);
+            });
 
-            vec![text_node]
+            vec![self.text_node.unwrap()]
```
MinusGix, line 729: Minor: can just use the `text_node` from a few lines ago

src/views/text_input.rs
```diff
@@ -947,426 +627,221 @@ impl Widget for TextInput {
...
-        self.placeholder_style.read_style(cx, &placeholder_style);
+        // To set wrap method to `None`
+        self.editor.with_untracked(|ed| {
+            ed.es.update(|s| {
+                if s.read(cx) {
+                    ed.floem_style_id.update(|val| *val += 1);
+                    cx.app_state_mut().request_paint(id);
+                }
+            })
+        });
```
MinusGix, line 692: I'm confused what this has to do with setting wrap to `None`

MinusGix:
Overall it looks good at a skim.  
It would be possible/better to have it more integrated with the existing editor logic, it could be implemented in terms of `EditorView` with some extensions/alterations to `Editor` and `EditorView`; but this method does avoid some extra work that the full `EditorView` would have needed, which is admittedly nice. So I think it is good, even if I'd somewhat prefer to avoid duplication of selection/click/otherbits.

src/views/text_input.rs
```diff
@@ -59,872 +67,545 @@ prop_extractor! {
...
+pub fn text_input(text: impl Fn() -> String + 'static) -> TextInput {
+    let id = Id::next();
 
-#[derive(Clone, Copy, Debug)]
-pub enum Direction {
-    Left,
-    Right,
-}
+    let text_editor = text_editor(text());
+    let editor = create_rw_signal(text_editor.editor().to_owned());
```
MinusGix, line 77: You don't need to create a `TextEditor` instance to get an `Editor`. Just use the same logic that `text_editor` uses to create it.

src/views/text_input.rs
presiyan-ivanov: I do want  to keep the note on update order, but I can remove the 2nd line

src/views/text_input.rs
```diff
@@ -59,872 +67,545 @@ prop_extractor! {
...
         data: ViewData::new(id),
-        cursor_glyph_idx: 0,
-        placeholder_text: None,
-        placeholder_buff: None,
-        placeholder_style: Default::default(),
-        selection_style: Default::default(),
-        buffer,
-        text_buf: None,
+        editor,
+        offset: 0,
```
MinusGix, line 191: Not sure you need the `offset` field, can't you just use the editor's cursor offset?

src/views/text_input.rs
```diff
@@ -59,872 +67,545 @@ prop_extractor! {
...
+        preedit_range: None,
+        layout_rect: Rect::ZERO,
+        content: "".to_string(),
```
MinusGix, line 194: Why is this separate from the editor's Rope?  
(Is it to more efficiently create the text layout? Or?)

examples/widget-gallery/src/clipboard.rs
presiyan-ivanov: I've moved it  to the widget fn(since an input without `keyboard_navigatable` is not really functional), so user code does not have to specify it.

MinusGix:
Also, I think it'd be nice to have a text input func for 'arbitrary function which you then have to hook `on_update` to do anything with' and a text input func for 'uses & updates `RwSignal<String>`' like the previous.   
Personally, I'd be inclined to call the input taking a function as `text_input_fn` and have the one taking `RwSignal<String>` as the 'default' `text_input`. That will make common shorthand easier. @dzhou121 if you have any thoughts on naming/utility functions for the view.

dzhou121:
I feel like text_input can be a normal editor without new lines? 

So maybe it makes more sense to add this to the editor widget?

presiyan-ivanov:
> Also, I think it'd be nice to have a text input func for 'arbitrary function which you then have to hook `on_update` to do anything with' and a text input func for 'uses & updates `RwSignal<String>`' like the previous. Personally, I'd be inclined to call the input taking a function as `text_input_fn` and have the one taking `RwSignal<String>` as the 'default' `text_input`. That will make common shorthand easier. @dzhou121 if you have any thoughts on naming/utility functions for the view.

i was thinking the same actually(I had this, but decided to not include it). We can add something like  `rw_text_input`

src/views/text_input.rs
presiyan-ivanov: Well usually the *text_editor* view reads  EditorStyle, but since I am not using it, I need to do the same in the input to get it

presiyan-ivanov:
> dzhou is essentially proposing what I was saying before, so that it is just the typical `EditorView` rather than doing custom drawing/events.

Ah.. now I understand. Sounds reasonable, yeah.

PR #424, opened 2024-04-21
Text rotation

15artis:


15artis:
Need to add add_xform() to render_glyph

lewis-carson:
Is there anything else needs implemented? I can lend a hand but this looks pretty feature complete to me

.vscode/settings.json
panekj, line 1: Delete that file

src/style.rs
```diff
@@ -1020,6 +1021,14 @@ pub enum TextOverflow {
...
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Rotation {
+    Rotation0,
+    Rotation90,
+    Rotation180,
+    Rotation270,
+}
+
```
panekj, line 1031: This should have better naming like:
```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum Rotation {
    NoRotation,
    RotateRight,
    RotateUpsideDown,
    RotateLeft,
}
```

src/style.rs
```diff
@@ -1716,6 +1736,19 @@ impl Style {
...
+    pub fn rotation_90(self) -> Self {
+        self.rotation(Rotation::Rotation90)
+    }
+    pub fn rotation_180(self) -> Self {
+        self.rotation(Rotation::Rotation180)
+    }
+    pub fn rotation_270(self) -> Self {
+        self.rotation(Rotation::Rotation270)
+    }
+
```
panekj, line 1751: same as above

panekj (changes requested):

src/style.rs
panekj: although I don't feel like `rotate` is anyhow good

panekj: Preferably it should somehow comply with how CSS defines this attributes

15artis: ok, thanks for your advice. I will revise it right away

dzhou121:
Does this require [the vger RP](https://github.com/lapce/vger-rs/pull/14)?

15artis:
yes

PR #432, opened 2024-04-30
ECS

dzhou121:
This PR adds `VIEW_STORAGE` which stores all the `View`, `ViewState`, parent/child relationship between views, and `TaffyTree`.  The benefits: 

- No need for `ViewData` any more. `ViewData` was introduced so that it can store the initial `Style`, `EventListener` without go through the `UpdateMessages`. Now we have "direct" access to `ViewState` so this is not needed.
- Lots of other updates don't need to go through the `UpdateMessages` 
- `update_state` doesn't need to go through the view tree by id_path dispatch, since we have "direct" access to `View`
- The taffy layout of an view can be accessed anywhere without the need of `AppState`


jrmoulton:
For what I care about this looks great to me. Removes the boilerplate for view creation and no longer has duplicated locations for events/styles. 👍

dzhou121:
Some exciting updates. 

`IntoView` is re-purposed to no need for a view id, which means we can impl it for &str, (), etc. And it will work like a normal `View` so can directly apply `Decorators` on it. E.g. you can do 
```rust
(
    "some button",
   (    
        ().style(|s| s.width(100).height(20).background(Color::GRAY),
        "another button".style(|s| s.color(Color::RED)),
   )
).style(|s| s.flex_col())
```

jrmoulton:
I definite like Into view for stacks. I'm on the fence though whether we should implement it for the empty tuple and`&str`. 

I feel like `text`/`empty` make it significantly more clear what is going on. 

I'd be fine with it though / maybe would come to love it

dzhou121:
Added `impl IntoView` for `Box<dyn Fn() -> IntoView>` and `RwSignal<IntoView>`

PR #433, opened 2024-05-01
Editor invalidation, optimizations, and more

MinusGix:
This does a variety of optimizations.  
Commits:  
- Rather than (roughly) `HashMap<Font Size, HashMap<Line, TextLayout>>`, we store it as (roughly) `HashMap<Font Size, [TextLayout]>`.
  - Also introduces smarter invalidation of text layouts, using `InvalLines`, rather than the recreating the text layouts after *every edit*
  - This makes a lot of other changes simpler, too.
- LLVM seems pretty poor at optimizing Rope. This results in many cases where the same value is computed repeatedly, and so I've started adding more utility functions. I've also cached the `last_line` of the Rope on the `RopeTextVal`/`RopeTextRef`, as this gives some boost due to how often we compute it.
  - I'd like to cache this on the `Rope` itself, but it'd require a good investment of time to wrap my head around all the implementation details of `Rope`. (which is also why some other opts aren't done)
- Due to the new layout cache as an array, calculating the last visual line becomes significantly cheaper, as it is easy to only calculate the vline part that is actually in view.
- This and the next commit make large text in register for pasting faster.
- Fix the cache, there was a bug with how I invalidated
- Apparently I swapped the order of the 'forward' / 'backward' search which made it slower than needed, but these are removed later
- Removes the `HashMap<Font Size, _>` as it complicates a bunch of code having multiple locations for the layouts to come from.
  - This does mean less caching for features like code lens in Lapce (or whatever it was renamed to) where some text is made very small, but I think we are fast enough at creating text layouts for that to *fine*.
  - This gives a very nice performance boost in some areas, removing the majority of lag large files have
- Completely remove font size provider. This wasn't needed anymore, really, since `Lines` doesn't need to know the font size. This simplifies the code.
- Next three commits are simply altering some functions to avoid the possibility of allocating a string during operations that we might use often. These aren't significant, though.
- Another two opts because LLVM is being mean
- Simplifies the calculation of the initial vline like how last vline was made cheaper, making calculating the visual line practically free (especially compared to the cost of creating text layouts)
- Fix invalidation logic again, and fix it across multiple editors
- Make the standard `InvalCount` not need the rope as the majority of locations do not benefit from it, and it would make simply invalidating some lines harder. Still has the version with `Rope` separate as `InvalCountR`
- Remove `cache_rev` from `Document` (and the impl supplied, `TextDocument`), as well as from Visual Lines, as we no longer need an entire file cache id due to line-specific invalidation. (All the lines can still be cleared by calling a specific function on `Lines`, possibly should be exposed from `Editor` if it isn't already)
- Make the `Layouts` cache a generic `LineRenderCache` which can be used in Lapce for various data associated with lines that needs to be invalidated

There's still a few bugs I need to determine, and may try other alterations.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,6 +68,9 @@ image = { workspace = true }
 im = { workspace = true }
 wgpu = { workspace = true }
 
+[dev-dependencies]
+criterion = "0.5"
+
 [features]
 default = ["editor", "default-image-formats"]
 # TODO: this is only winit and the editor serde, there are other dependencies that still depend on
@@ -99,3 +102,10 @@ tokio = ["dep:tokio"]
 # rfd (file dialog) async runtime
 rfd-async-std = ["dep:rfd", "rfd/async-std"]
 rfd-tokio = ["dep:rfd", "rfd/tokio"]
+
+[profile.bench]
+debug = true
+
+[[bench]]
+name = "basic_editing"
+harness = false


charlescgs:
@MinusGix Do you have any plans to continue to work on this PR?

MinusGix:
No, I have mostly stalled on contributing to Floem/Lapce. More time requirements on me and somewhat burnt on it.
These changes would still be quite good to have, though there was questions about implementing it in Lapce. So anyone can feel free to rework it or whatever.
(Last I saw Lapce's semantic highlighting is requested in full every time. This makes the faster partial updates this PR provides not as useful due to a full update. Lapce would benefit from semantic highlighting getting only deltas but that requires some redesign work because Lapce's proxy doesn't give the raw values from the LSP to the frontend.)

umbrageodotus:
Should I work on it?

PR #438, opened 2024-05-11
Do not jump to end of text while editing (fixes #282)

### LINKED ISSUE #282: Text input cursor jumps to end when inserting/removing character within the string
https://github.com/lapce/floem/assets/1407980/cddb5665-e852-451f-aef0-084f44b6af04

This line appears to trigger, causing the cursor glyph index to jump to the end of the line:
https://github.com/lapce/floem/blob/main/src/views/text_input.rs#L948

Something in 189985f459f1533399e1b5e26cf95f63ea19dbb9 appears to have triggered a bit of regression.

dwuertz:
Only move cursor index to end of buffer when focus is gained


jrmoulton:
Hi @dwuertz, this issue ended up being fixed in another PR along with several other fixes to the text input. I'll go ahead and close this as it's been fixed elsewhere but thanks for taking the time to work on the issue

PR #453, opened 2024-05-22
Expose setting a Window Icon

thewizzy:
The ability to set a window icon is available in floem-winit but it isn't exposed through floem. This allows the floem dev to set a window icon as part of the window configuration.

There is an example included and it borrows heavily from the [floem winit window_icon example](https://github.com/lapce/winit/blob/master/examples/window_icon.rs) - including the same icon file.


src/window.rs
```diff
@@ -17,6 +18,7 @@ pub struct WindowConfig {
     pub(crate) transparent: Option<bool>,
     pub(crate) fullscreen: Option<Fullscreen>,
     pub(crate) window_icon: Option<bool>,
+    pub(crate) window_icon_value: Option<Icon>,
```
thewizzy, line 21: I think we could change the previous line `window_icon` to be an Option<Icon> instead and remove `window_icon_value`.
This is a breaking change to the API - so the choice is for floem devs.

dzhou121: I think we can re-use `window_icon` because it's not used anywhere

panekj:
It should be noted from where the icon comes from and what licence it is under

thewizzy:
> It should be noted from where the icon comes from and what licence it is under or the icon should be replaced with one that doesn't require attribution/licence (for example Ferris could be used like in #116)

Absolutely. I have updated it to use a copy of existing assets that you mentioned. Thank you.

src/window.rs
thewizzy: This has been changed. It is much nicer this way, thank you for the agreement.

PR #456, opened 2024-05-22
Support undecorated windows in `WindowConfig`

timboudreau:
Rationale: Palette style windows are useful for components beyond menus and combo boxes.  This
patch adds support for undecorated windows in WindowConfig.

Use-case: I have a fairly large popup I need to show to select something where a lot of
visual detail needs to be visible to the user; while I have this implemented with `add_overlay()`,
I do not expect that to be usable in a pleasant way if the window size is smaller than what the
popup requires (scrolling would be an ugly solution).

Built and tested on Mac OS and Gentoo Linux without problems.

### DIFF
--- a/src/app_handle.rs
+++ b/src/app_handle.rs
@@ -279,6 +279,19 @@ impl ApplicationHandle {
                     window_builder = window_builder.with_decorations(false);
                 }
             }
+            if let Some(undecorated) = config.undecorated {
+                window_builder = window_builder.with_decorations(!undecorated);
+                #[cfg(target_os = "macos")]
+                if undecorated {
+                    use floem_winit::platform::macos::WindowBuilderExtMacOS;
+                    // A palette-style window that will only obtain window focus but
+                    // not actually propagate the first mouse click it receives is
+                    // very unlikely to be expected behavior - these typically are
+                    // used for something that offers a quick choice and are closed
+                    // in a single pointer gesture.
+                    window_builder = window_builder.with_accepts_first_mouse(true);
+                }
+            }
             if let Some(transparent) = config.transparent {
                 window_builder = window_builder.with_transparent(transparent);
             }
--- a/src/id.rs
+++ b/src/id.rs
@@ -84,7 +84,7 @@ impl ViewId {
         })
     }
 
-    pub(crate) fn view(&self) -> Rc<RefCell<Box<dyn View>>> {
+    pub fn view(&self) -> Rc<RefCell<Box<dyn View>>> {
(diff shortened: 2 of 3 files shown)


dzhou121:
The diff doesn't look right here. Maybe needs a cleanup?

timboudreau:
I did another pr that was merged which added a different field on the same line, so probably I need to rebase this patch.

timboudreau:
Okay, PR is rebased from main and I moved the field and method declarations to keep them from colliding with #457 - will now check that one for conflicts as well.

src/app_handle.rs
```diff
@@ -279,6 +279,19 @@ impl ApplicationHandle {
                     window_builder = window_builder.with_decorations(false);
                 }
             }
+            if let Some(undecorated) = config.undecorated {
+                window_builder = window_builder.with_decorations(!undecorated);
```
panekj, line 283: unnecessary double negation, this should not be `!` and called `undecorated`

panekj: > Every toolkit I've worked with for years has referred to the special-case of windows with no frame or title-bar as "undecorated".

We're not every toolkit and how lack of decorations is referred to doesn't have to dictate how API looks like.

> Consider that if we called this property "decorated" and had a method with that name, what the person trying to call the method is going to perceive is that undecorated windows are the default (they aren't and shouldn't be), and they must call decorated(true) to avoid that.

By that logic `show_titlebar` or other options are also `false` by default.
And while being on topic of `show_titlebar`, I find the `undecorated` option to be redundant and harmful given that we already configure decorations in the titlebar option. This should be adjusted or consolidated.

timboudreau: *Undecorated* means something different.

On most desktop OS's, a window that had its titlebar hidden still has a system-drawn *frame* that may or may not allow the window to be moved or resized.

*Undecorated* means something different - the application is responsible for painting the **entire** visual representation of the window, and the system paints nothing at all.

It is a genuinely different thing than hiding the titlebar, and has been used for things like popups and shaped, non-rectangular windows since the 90s.

Make just one - either one - available, and there are some applications you should be able to write with Floem but you can't.

timboudreau: And, show_titlebar **should** be hide_titlebar - the caller has expectations of what *make me a window* means, trained by decades of desktop UI APIs, and the API should express the way you can ask that the window *deviate* from the platform default.

But I'm trying to flesh out missing functionality, not make gratuitous breaking changes in an existing API.

PR #457, opened 2024-05-22
Add the ability to configure mac-os specific properties to WindowConfig

timboudreau:
Rationale: Mac OS supports a number of window features that users of Mac OS applications expect, but which don't make sense as lowest-common-denominator cases.

I am one of the authors of NetBeans, and did *all* of the initial *make it Mac user friendly* circa 2002, which was extra difficult due to Java AWT's purity-test that any windowing feature that was not supported by Window, Solaris, Linux *and* Mac OS had no business being supported by AWT.  That significantly worsened the user experience (eventually Apple supported magic "component properties" that made some improvements possible), and is not a mistake you want to imitate.

I've endeavored to make the result minimally invasive to the existing API - all Mac OS specific properties are on a separate `struct` accessed via a closure, but did ensure that the types are *visible* on other OS's, so developers creating floem-based apps do not need to litter their code with `#[cfg(target_os = "macos")]`.

Use Case: I have some palette windows that I need more control of the appearance of than is possible without this.

Built and tested on Mac OS *and* Gentoo Linux without problems.

### DIFF
--- a/src/app_handle.rs
+++ b/src/app_handle.rs
@@ -294,6 +294,43 @@ impl ApplicationHandle {
             if let Some(window_icon) = config.window_icon {
                 window_builder = window_builder.with_window_icon(Some(window_icon));
             }
+            #[cfg(target_os = "macos")]
+            if let Some(mac) = config.mac_os_config {
+                use floem_winit::platform::macos::WindowBuilderExtMacOS;
+                if let Some(val) = mac.movable_by_window_background {
+                    window_builder = window_builder.with_movable_by_window_background(val);
+                }
+                if let Some(val) = mac.titlebar_transparent {
+                    window_builder = window_builder.with_titlebar_transparent(val);
+                }
+                if let Some(val) = mac.titlebar_hidden {
+                    window_builder = window_builder.with_titlebar_hidden(val);
+                }
+                if let Some(val) = mac.full_size_content_view {
+                    window_builder = window_builder.with_fullsize_content_view(val);
+                }
+                if let Some(val) = mac.movable {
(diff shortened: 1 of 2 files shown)


src/app_handle.rs
```diff
@@ -294,6 +294,43 @@ impl ApplicationHandle {
             if let Some(window_icon) = config.window_icon {
                 window_builder = window_builder.with_window_icon(Some(window_icon));
             }
+            #[cfg(target_os = "macos")]
+            if let Some(mac) = config.mac_os_config {
+                use floem_winit::platform::macos::WindowBuilderExtMacOS;
+                if let Some(val) = mac.movable_by_window_background {
+                    window_builder = window_builder.with_movable_by_window_background(val);
```
ProgramCrafter, line 301: Is there an option to make this something like `window_builder.opt_change_movable_by_window_background(&mac.movable_by_window_background);`? Not in this PR if there is need to minimize diff, but generally.

timboudreau: I don't really understand what you're asking here.

You want to change the name of something? The character case? Or the functionality?

timboudreau: Okay, seeing this on the computer I think I see what you mean.  `WindowBuilder` is part of [winit](https://github.com/rust-windowing/winit.git) - it is just used (and exported) by Floem.  The method name you want to change is not part of this project.

dominikwilkowski:
This is such a great PR. There are some macos specific bugs that need attention here including supporting [`REOPEN`](https://github.com/lapce/winit/pull/5). Appreciate the focus!

PR #461, opened 2024-05-23
deps: kurbo is used via peniko.

waywardmonkeys:
Since all usages of `kurbo` are via `peniko`, the direct dependency on `kurbo` is not needed. This makes it easier to keep versions in sync.

The re-export of `kurbo` was happening directly so it could have been out of sync with an updated `peniko`, so re-export the version from `peniko` instead.


waywardmonkeys:
(Noticed this when looking at your branch that replicates the `vello` bug.)

dzhou121:
peniko 0.1.0 doesn't give the options to turn on serde feature on kurbo. That's why we had to use kurbo directly.

waywardmonkeys:
I see.

The current `main` branch of `peniko` has `serde` support now, so would it be useful if we did a release of that?

dzhou121:
> I see.
> 
> The current `main` branch of `peniko` has `serde` support now, so would it be useful if we did a release of that?

Yes!

waywardmonkeys:
Okay, `peniko` version 0.1.1 is published now and this PR is now updated to use it and enable the `serde` feature.

dzhou121:
Thanks!

PR #464, opened 2024-05-26
Fix #463 - broken inter-window update messages

### LINKED ISSUE #463: Inter-window messages sent to a view via `ViewId.update_state()` are broken
This one I'm not sure how to solve:  If I have a view in a popup window, and it sends a message to the view that created it in the main application window, the message is delivered (and the message may close the popup!) - but the view does not process the message until some OS-level event like waving the mouse over the main window causes it to wake up and process messages in its queue.

Here is a gist with a [complete, runnable example](https://gist.github.com/timboudreau/5e5d6145290de0b1446d5aa690ba488e) demonstrating the problem - I'd recommend just copy/pasting it and running it (instructions on exactly what to do will be in the main window) - it's big as bug demos go, but I figured it would help to have something that shows the utility of this sort of thing.

I get the optimization I think is being attempted here - if the window is not active, have it be *completely* quiet.  In general, that's a good thing - but this is a little too aggressive.  If it's intended to remain that way, then there needs to be some way for a caller to flag a message as *urgent* or at least a way to bang on its event loop and force some messages through.

timboudreau:
This addresses a couple of issues:

 * Messages (including repaints) not getting processed when delivered a View in a window which is inactive
 * Ability to retrieve the window id of a view without the application having to pass that information down to whatever code needs it

and addresses a related issue (related to dealing with WindowIds - the solution falls out of the solution for the above):

 * Ability to retrieve the screen position of the window if you have its id

This patch adds atracking data structure:

 * A static mapping of `root-id:WindowId`:`Arc<Window>` which is populated on window creation and cleared on destruction, which makes it possible to access `Window.request_redraw()` to force a redraw.  It is guarded by a read-write lock (but written to only during window creation and destruction).

So, a ViewId that wants its window-id looks up its root view's id, and looks up the WindowId associated with that root.

We extend `WindowId` with methods to get the window bounds on screen with and without decorations and similar - these methods could be added to `ViewId` instead, but it seems more intuitive that you ask a window id for the window's bounds the same way you ask a view id for the view's bounds - that seems like what users will expect.

Given the usage patterns (the mapping will only be locked for writing when a window is being created or destroyed, and will not be read-locked except in the case of forcing a repaint or looking up window coordinates), it is not touched in the critical path of rendering - unless the application and destroying many windows at the same time on different threads, I would not expect the read-write lock to be contended at all in practice (famous last words, I know).

A thread-local for the mapping, as is done in `update.rs` might work (or look like it works), but we would be assuming no platform floem will ever support uses different event threads for different windows' event loops. I know of two that do - BeOS and Java AWT (with its stack of event threads to allow modal dialogs that work when their parent is disabled or blocked, which was a source of "interesting" deadlocks).

If there was some path that I missed that would let a `ViewId` reach down to the `WindowHandle` that owns it without doing this sort of tracking, I would love to know about it - `PaintCx` has this information, but it is the only thing that does, and the `ViewId` has no access to it.


src/id.rs
```diff
@@ -252,8 +258,34 @@ impl ViewId {
...
+
+    pub fn window_id(&self) -> Option<WindowId> {
+        self.root().and_then(window_id_for_root)
+    }
+
     pub fn request_paint(&self) {
+        let active = self.is_in_active_window();
         self.add_update_message(UpdateMessage::RequestPaint);
+        if !active {
+            self.force_repaint();
```
dzhou121, line 287: do we still need this given that we've now processing all update messages, and if there's a repaint needed, the repaint request would be sent for that window?

timboudreau:
This PR is superseded by [this one](https://github.com/lapce/floem/pull/473) (already integrated) and [this one](https://github.com/lapce/floem/pull/480).  Closing.

PR #482, opened 2024-05-30
Fix dyn container

jrmoulton:
I'm not sure why this is different than before but this matches dyn_view and no longer panics


dzhou121:
The issue is that there’s no self.child_id = new_child_id in update

jrmoulton:
> The issue is that there’s no self.child_id = new_child_id in update

🤦

PR #486, opened 2024-06-03
Expose `View::default_compute_layout`

timboudreau:
While this method *is* implementable in custom component by copy-paste programming, it would be preferable for custom components to be able to keep up with any changes in the default implementation by calling it directly instead.


dzhou121:
maybe we can put the method under `ComputeLayoutCx` and call it `compute_view_children_layout`?

timboudreau:
Sure, sounds good.

PR #489, opened 2024-06-05
Fix focus handling in multi-window applications

timboudreau:
This took some time to debug - starting with an application that opens a popup window,
wherein the user should be able to tab between items and press a key to select and close
the popup.

It turned out to be **three** separate issues:

1. `window_handle.process_central_messages()` would discard any messages for `View`s with a different root than its own, so they would never be processed.
2. `ViewId` attempted to cache the root view id, but would just cache the deepest parent found - so if you do something that generates an update message
during construction of a tree of views, the cached value would not actually be the root view for the (not yet existing) window
3. Not quite as strictly a bug, but I can't imagine why it would be by design: `FocusGained` and `FocusLost` events caused by a call to `ViewId.request_focus()`
*would* be delivered to listeners, but a `View` would never receive them via `event_before_children()` or `event_after_children()` - so this one had a
workaround, but the workaround - a view attaching listeners to itself - would be much less readable than a straightforward implementation of one of
the event handler methods

The fix for 3. I am the least sure about:

* It is a bit ugly (swapping an `EventCx` for an `UpdateCx` to call `unconditional_view_event()` and then recreating the `UpdateCx`)
* It is not clear if there are *other* kinds of events which have the same problem of non-delivery to the view itself
* `ViewId.apply_event()` might be a cleaner place to implement it (though it would mean a bunch of calling back and forth between `WindowHandle` and `ViewId` which might be less than ideal)
* I do *not* stop propagation of events if `event_*_children()` returns `EventPropagation::Stop` - as long is this is about focus events, I think it is probably harmful to be able to block propagation of focus events **at all** - window focus events more so, since it's unlikely that one view in a tree knows for sure that none of its siblings need to know about the window losing focus - but the same logic applies more weakly to plain view focus events - it's a recipe for hard-to-diagnose focus bugs.

Since I'm a proud member of the 1990's `println` school of debugging, `UpdateMessage` now implements `Debug` - needed it to prove that retained messages
really were eventually processed, and it will probably be useful in the future.

### DIFF
--- a/src/view_storage.rs
+++ b/src/view_storage.rs
@@ -52,18 +52,18 @@ impl ViewStorage {
         }
     }
 
+    /// Returns the deepest view ID encountered traversing parents.  It does *not* guarantee
+    /// that it is a real window root; any caller should perform the same test
+    /// of `window_tracking::is_known_root()` that `ViewId.root()` does before
+    /// assuming the returned value is really a window root.
     pub(crate) fn root_view_id(&self, id: ViewId) -> Option<ViewId> {
-        let mut parent = self.parent.get(id).cloned().flatten();
-        while let Some(parent_id) = parent {
-            match self.parent.get(parent_id).cloned().flatten() {
-                Some(id) => {
-                    parent = Some(id);
-                }
-                None => {
-                    return parent;
-                }
-            }
+        // Rewritten from the original looping version - unless we anticipate view ids
+        // deeper than the possible stack depth, this should be more efficient and
+        // require less cloning.
+        if let Some(p) = self.parent.get(id).unwrap_or(&None) {
+            self.root_view_id(*p)
+        } else {
+            Some(id)
         }
-        parent
     }
 }
--- a/src/window_handle.rs
+++ b/src/window_handle.rs
@@ -310,6 +310,9 @@ impl WindowHandle {
             }
         }
         if was_focused != cx.app_state.focus {
+            // What is this for? If you call ViewId.request_focus(), this
+            // causes focus to be set to your view, then briefly set to
+            // None here and then back.  Why?
             cx.app_state.focus_changed(was_focused, cx.app_state.focus);
         }
 
@@ -701,11 +704,28 @@ impl WindowHandle {
         CENTRAL_UPDATE_MESSAGES.with_borrow_mut(|central_msgs| {
             if !central_msgs.is_empty() {
                 UPDATE_MESSAGES.with_borrow_mut(|msgs| {
-                    let central_msgs = std::mem::take(&mut *central_msgs);
-                    for (id, msg) in central_msgs {
+                    // We need to retain any messages which are for a view that either belongs
+                    // to a different window, or which does not yet have a root
+                    let removed_central_msgs =
+                        std::mem::replace(central_msgs, Vec::with_capacity(central_msgs.len()));
+                    for (id, msg) in removed_central_msgs {
                         if let Some(root) = id.root() {
                             let msgs = msgs.entry(root).or_default();
                             msgs.push(msg);
+                        } else {
+                            // Messages that are not for our root get put back - they may
+                            // belong to another window, or may be construction-time messages
+                            // for a View that does not yet have a window but will momentarily.
+                            //
+                            // Note that if there is a plethora of events for ids which were created
+                            // but never assigned to any view, they will probably pile up in here,
+                            // and if that becomes a real problem, we may want a garbage collection
+                            // mechanism, or give every message a max-touch-count and discard it
+                            // if it survives too many iterations through here. Unclear if there
+                            // are real-world app development patterns where that could actually be
+                            // an issue. Since any such mechanism would have some overhead, there
+                            // should be a proven need before building one.
+                            central_msgs.push((id, msg));
                         }
                     }
                 });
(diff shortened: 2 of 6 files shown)


timboudreau:
Ignore the fact that [Expose View::default_compute_layout](https://github.com/lapce/floem/pull/489/commits/57adef1fd4e91736732d9b4313e5502f43ee3fc3) appears to be one of the commits here - I did this on my branch containing that commit, but the commit that applies here reverts it.

timboudreau:
Noted in comments, but will mention this here:  Retaining messages for unrecognized view roots has the potential to leak `ViewId`s **if** it there is any situation where it is normal to publish messages to a `ViewId` that either will never be owned by a window root, or which was but the window was closed before the message was encountered.  I.e. if you just called `ViewId::new().request_focus()` you would add a focus request *permanently* to `CENTRAL_UPDATE_MESSAGES`.

You know more about the common use-cases for Floem than I do - is this likely to be a problem?

If so, the simplest gc-like solution would just be to wrap messages in a struct that keeps a counter, increment the counter each time the same message is seen in the queue but not processed, and if it crosses some magic threshold number, don't put it back.  Any other solution would involve diffing collections and could get expensive, while if a message hangs out in the queue unprocessed for 100 events (or whatever number seems right), that's not going to have a big impract.

I think probably only implement a solution if we find out that it's actually a problem - fiddling with `ViewId`s and abandoning them is surely an application-level bug for the most part, although I could imagine a minor case or two where a view races with window removal (like, a click action dispatches `request_focus` but something else intercepts the click and decides to close the window).

timboudreau:
Just in case you're wondering what on earth I'm doing that I'm running into these sorts of issues, here's a screen shot of a prototype of the DSP plugin UI:

(Actually plumbing this thing, given need to interoperate with pointers handed to me over FFI as if they were `RwSignal` is a remaining challenge)
Screenshot 2024-06-05 at 7 36 27 PM

jrmoulton:
> 3. Not quite as strictly a bug, but I can't imagine why it would be by design: FocusGained and FocusLost events caused by a call to ViewId.request_focus()would be delivered to listeners, but a View would never receive them via event_before_children() or event_after_children()

I think is a regression after the ECS pr

src/view_storage.rs
```diff
@@ -52,18 +52,18 @@ impl ViewStorage {
...
-                    return parent;
-                }
-            }
+        // Rewritten from the original looping version - unless we anticipate view ids
+        // deeper than the possible stack depth, this should be more efficient and
+        // require less cloning.
+        if let Some(p) = self.parent.get(id).unwrap_or(&None) {
+            self.root_view_id(*p)
+        } else {
+            Some(id)
```
jrmoulton, line 66: Even though it's unlikely I don't think we are better off using recursion here. That will place a (unlikely) limit on the max depth. 

Cloning the ViewId should be find. It's only 8 bytes and copies should just all be kept in registers.

src/window_handle.rs
```diff
@@ -310,6 +310,9 @@ impl WindowHandle {
             }
         }
         if was_focused != cx.app_state.focus {
+            // What is this for? If you call ViewId.request_focus(), this
+            // causes focus to be set to your view, then briefly set to
+            // None here and then back.  Why?
```
jrmoulton, line 315: the app_state focus will taken from it's option before a pointer down even is propagated. Once event propagation is done, if another view has set itself to have focus because of the pointer down event the focus_changed function will notify the old was_focused that it's focus has been lost and the new app state focus that it has gained focus

src/view_storage.rs
timboudreau: Happy to leave it as-is too.  Though I think if you have enough view ids to exhaust the stack, you might have some UI usability problems far worse than hitting that limit :-)

src/window_handle.rs
timboudreau: What I found (I added dumping the stack to the focus method it calls if the value was `None`) is that, when `someViewId.request_focus()` was called with an existing focus target, focus goes old_id -> None -> new_id.

I didn't check if events were fired for focus going to `None`, since event propagation wasn't working at that point.  As long as we're not doubling up event notifications, it's probably fine - just peculiar.

timboudreau:
> > 3. Not quite as strictly a bug, but I can't imagine why it would be by design: FocusGained and FocusLost events caused by a call to ViewId.request_focus()would be delivered to listeners, but a View would never receive them via event_before_children() or event_after_children()
> 
> I think is a regression after the ECS pr

Not familiar with that PR.  Is there a less invasive (or at least less clunky) fix, then?

src/window_handle.rs
```diff
@@ -748,11 +774,40 @@ impl WindowHandle {
                         if cx.app_state.focus != Some(id) {
                             let old = cx.app_state.focus;
                             cx.app_state.focus = Some(id);
+
+                            // Fix: Focus events were never dispatched to views' own
+                            // event_before_children / event_after_children.
+                            //
+                            // This is a bit messy; could it be done in ViewId.apply_event()
+                            // instead?  Are there other cases of dropped events?
```
dzhou121, line 782: Other events are from winit, so they would be dispatched by the `event` method, while `Event::FocusGained` and `Event::FocusLost` is kind of an "internal" event.

src/window_handle.rs
```diff
@@ -748,11 +774,40 @@ impl WindowHandle {
...
+                            if let Some(old) = old.as_ref() {
+                                ec.unconditional_view_event(*old, Event::FocusLost, true);
+                            }
+                            ec.unconditional_view_event(id, Event::FocusGained, true);
+
+                            cx = UpdateCx {
+                                app_state: ec.app_state,
+                            };
+
                             cx.app_state.focus_changed(old, cx.app_state.focus);
```
dzhou121, line 796: `unconditional_view_event` probably needs to called in `focus_changed`.

E.g. when the focus got changed due to pointer click.

Issue #492, opened 2024-06-05
Text selection panic on multi-line labels


dominikwilkowski:
If I extend the widget gallery with the code below:

`widget-gallery/src/labels.rs`
```diff
use floem::{
    cosmic_text::{Style as FontStyle, Weight},
    peniko::Color,
    views::{label, static_label, tooltip, Decorators},
    IntoView,
};

use crate::form::{form, form_item};

pub fn label_view() -> impl IntoView {
    form({
        (
            form_item("Simple Label:".to_string(), 120.0, || {
                tooltip(label(move || "This is a simple label".to_owned()), || {
                    static_label("This is a tooltip for the label.")
                })
            }),
            form_item("Styled Label:".to_string(), 120.0, || {
                label(move || "This is a styled label".to_owned()).style(|s| {
                    s.background(Color::YELLOW)
                        .padding(10.0)
                        .color(Color::GREEN)
                        .font_weight(Weight::BOLD)
                        .font_style(FontStyle::Italic)
                        .font_size(24.0)
                })
            }),
+             form_item("Simple Label:".to_string(), 120.0, || {
+                 "My multiline label\nWith a second line\nAnd a third"
+             }),
        )
    })
}
```

The text select crashes as soon as I start the selection on the middle line and select up and down.
![text-selection](https://github.com/lapce/floem/assets/1266923/a898a38e-d005-4452-ab7e-3a8b90fee06e)

jrmoulton:
I have other multiline labels that are working just fine. Not sure why this one in particular is crashing

dominikwilkowski:
It crashes in my app as well which is how I found this.

dominikwilkowski:
fixed via #493

PR #493, opened 2024-06-05
remove debug log, always paint text selection

jrmoulton:


jrmoulton:
@dominikwilkowski I wasn't able to reproduce the crash but I've removed the debug log and fixed another thing. Can you see if this fixes the issue? And if not can you copy/paste the crash into the issue?

dominikwilkowski:
Can confirm that jrmoulton/floem at `4271b05d711b466fe154349177f2d673d2294a1e` fixes #492

PR #497, opened 2024-06-13
Scroll top, middle, and bottom

jrmoulton:
Support for scroll to center, middle and bottom while supporting scroll off.


jrmoulton:
The CI issues are unrelated to this PR. I think this is good to go

PR #499, opened 2024-06-20
tooltip improvements

jrmoulton:
Adds a tooltip trait and also extracts the tooltip style class from the context of the tooltip


jrmoulton:
I think this is good to go

PR #508, opened 2024-07-10
Convert examples/* to proper cargo-manifest structure.

maaku:
I hope this is well-received as I didn't ask before authoring. While evaluating floem for use in my own app, I started with the examples and got frustrated that I couldn't build them / run them / doc them from the workspace directory with the standard `cargo run --example <name>` command, or to build/test them all with e.g. `cargo build --examples`. This is because they are created as workspace crates, not as proper cargo examples.

This PR converts them all to standard cargo manifest examples.


panekj:
You should run them as `cargo run -p <name>`

timsueberkrueb:
Whether or not the structural changes will be merged, I think the added doc comments are helpful.

jrmoulton:
I'm going to go ahead and close this for now as I think the current structure is preferred but thanks for taking the time to work on improving things!

PR #510, opened 2024-07-16
inspector: Add a search box and select up and down

jm-observer:
Please let me know if it doesn't fit to merge

The effect is as follows:

https://github.com/user-attachments/assets/40ca4c22-b468-49f5-989b-156e5a99edd2


dzhou121:
This looks very useful. Can you please explain how the search works?

jm-observer:
Just go through all the views, match the names, then focus

PR #512, opened 2024-07-17
Make apply classes from context public

jrmoulton:
This function is useful for overlays to get all of the current styles on classes without the current style as used in dropdown


jrmoulton:
Actually I think just directly using the indirect style is what I am looking for

PR #514, opened 2024-07-17
re-request scroll style if a refresh is needed

jrmoulton:


dzhou121:
When does this needed? I'm a bit worried that this could cause infinite loop in unforeseen rare cases.

jrmoulton:
Hmm yeah maybe this isn't the best way to fix the issue I was having. 

```rust
root_view().style(|s| s.class(DropDownClass, |s| s.class(|s| s.class(ScrollClass, |s| s.background(BLUE)))

```

I have an issue where when the app starts, before user interaction, scroll views that are not children of DropDownClass will have the blue background applied. After clicking the blue background it then resolves to the correct styles. Moving the mouse hover in and out of that scroll view will sporadically cause it to switch to the blue background again until it is clicked (or some other interaction). Adding this change fixes that issue but maybe the actual issue is somewhere else.

PR #519, opened 2024-07-19
update wgpu

panekj:


panekj:
ok, nevermind, there is patch release for 0.19 and I don't care enough to go through updating vger and other things

PR #524, opened 2024-07-28
support WindowEvent::DroppedFile

jm-observer:


src/window_handle.rs
```diff
@@ -390,7 +394,17 @@ impl WindowHandle {
...
+        self.dropper_file = Some(path.clone());
+        self.event(Event::DroppedFile(DroppedFileEvent { path }));
+    }
+
     pub(crate) fn pointer_move(&mut self, pos: Point) {
+        if let Some(path) = self.dropper_file.take() {
+            self.event(Event::DroppedFileWithPosition(
+                DroppedFileWithPositionEvent { path, pos },
+            ));
+        }
```
dzhou121, line 407: The cursor position is stored in `self.cursor_position`. So we should be able to send the dropped file event with the cursor position when the dropped file event was trigger from winit.

jm-observer: I tried. It's not accurate. The mouse coordinates are not updated when you drag and drop

dzhou121: Yeah it's something at winit's end. It should be fixed when this is implemented https://github.com/rust-windowing/winit/issues/3833

Does it always emit pointer move after the file was dropped? If so, we don't need the drop file event without mouse position.

jm-observer: yes, my observation is this.

dzhou121: Shall we just have one event `DroppedFile` with both path and pos please?

jm-observer: Well, I just kept DroppedFileWithPositionEvent

dzhou121:
rename `DroppedFileWithPosition` to `DroppedFile` maybe?

PR #527, opened 2024-08-03
Add initial wasm support

timsueberkrueb:
Add initial support to run floem applications on the web using WASM and WebGPU. This only works on browsers that support WebGPU. WebGL2 via wgpu does not work because vger-rs uses storage buffers in its shaders which is not supported on WebGL2. Haven't looked into whether tinyskia can be made to work on the web.

Before this PR can be merged, the following changes need to land in `winit` and `cosmic-text` first:

* https://github.com/lapce/winit/pull/7
* https://github.com/lapce/cosmic-text/pull/7


dzhou121:
cosmic-text is now changed to the upstream version which doesn't have stretto

src/app_handle.rs
```diff
@@ -353,7 +392,34 @@ impl ApplicationHandle {
             return;
         };
         let window_id = window.id();
-        let window_handle = WindowHandle::new(window, view_fn, transparent, apply_default_theme);
+
+        #[cfg(target_arch = "wasm32")]
+        {
+            let canvas = window.canvas();
```
dzhou121, line 398: I'm wondering if it's better to provide winit with the canvas, so that we can rely on the browser to set the size of it.

timsueberkrueb: Done.

dzhou121:
Looks good to me now!

Although I tried to run webgpu example but my Chrome complains about the wgsl. Does it run properly for you?

timsueberkrueb:
Thanks for noticing, I can reproduce it. I had other problems with chrome before, so I mostly tested on Firefox nightly (on Linux). I'll look into these.

timsueberkrueb:
https://github.com/lapce/vger-rs/pull/16 fixes this problem for me.

timsueberkrueb:
Few things I would like to point out:

* [ ] Requires new floem-vger-rs release
* [ ] Requires new floem-winit release
* [ ] Example includes font files to the repository. Is this okay? Otherwise we can try to find another means to load these fonts (e.g. from some CDN).

PR #535, opened 2024-08-06
Add custom style trait

jrmoulton:
Makes it so that external crates can implement custom style updaters in an efficient way. Previous implementations in floem had access to private items. Now the default implementation of the trait has access to those items and provides the updater. 

This also makes the API for custom styles more consistent.

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -2,6 +2,7 @@
 //!
 //!
 
+use floem_reactive::create_updater;
 use floem_renderer::text::{LineHeightValue, Weight};
 use im_rc::hashmap::Entry;
 use peniko::Color;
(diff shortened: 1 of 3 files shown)


src/style.rs
```diff
@@ -1878,6 +1883,33 @@ impl Style {
     }
 }
 
+pub trait CustomStyle {
+    fn new_custom_style() -> Self;
```
dzhou121, line 1887: maybe we can lose this, and make `CustomStyle: Default`.

src/style.rs
```diff
@@ -1878,6 +1883,33 @@ impl Style {
...
+    fn get_style(&self) -> Style;
```
dzhou121, line 1888: maybe this can be `fn into_style(self) -> Style;`

jrmoulton:
I've removed the custom style trait and just used Default and Into

PR #536, opened 2024-08-06
Add conditional classes and debug names

jrmoulton:
This also reverses the display of class names in the inspector which makes classes applied at a higher level in the call stack appear first.


jrmoulton:
Any comments on this PR?

PR #541, opened 2024-08-07
support gradients/brush as background/border

jrmoulton:


jrmoulton:
I'm not sure if this should be merged or not. It only supports linear 2 stop gradients and not other gradients or images. 

It might be better to create a new brush type that doesn't allow an image to be used. 
I should probably also add support for the other gradient types as well first

dzhou121:
Thanks for adding this. I think we can add the missing parts later. It's already a big improvement.

PR #546, opened 2024-08-10
Remove the use of `LazyLock` in places where it's not necessary

sgued:
`Mutex::new` has been `const` on stable for a while so it can be used without the need for `Lazy`.

The use of `once_cell::Lazy` has also been replaced by the recently stabilized `std` version, that is already used somewhere else so there is no MSRV bump.

This PR also makes `ExtEventHandler` private (which is a breaking change), but this structure is only ever used in a private static and should  not have been made public from the start (unless we also make the static public).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,6 @@ serde = "1.0"
 lapce-xi-rope = { version = "0.3.2", features = ["serde"] }
 strum = { version = "0.26.2" }
 strum_macros = { version = "0.26.2" }
-once_cell = "1.17.1"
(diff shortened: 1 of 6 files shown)


panekj:
The MSRV still needs to be bumped, since std `LazyLock` is 1.80

sgued:
LazyLock was already used before  this PR in the renderer crate: https://github.com/lapce/floem/blob/908c3f32fab0e647f97db191e9bd79511fc16c64/renderer/src/text/layout.rs#L1

PR #548, opened 2024-08-11
Fix text input cursor jumping to the end

### LINKED ISSUE #532: Inserting text in the middle of a textbox is impossible
1. Open the Widget Gallery example
2. Choose "Input"
3. Type some text:
image
4. Move the cursor:
image
5. Attempt inserting more text in the middle:
image
6. The cursor jumps to the end of the textbox after the first keystroke.

sgued:
Fix #532 

See https://github.com/lapce/floem/issues/532#issuecomment-2282706511 for the explanation of what this PR fixes.

This PR fixes it by adding a "cache" of the previous state of the buffer that skips updating the cursor position if the buffer is the same after the update.

Ideally there should be a way to distinguish updates "from inside" and updates "from outside", so that the effect could simply not be run.

### DIFF
--- a/src/views/text_input.rs
+++ b/src/views/text_input.rs
@@ -270,7 +270,7 @@ impl TextInput {
                 false
             }),
             (Movement::Word, Direction::Left) if self.cursor_glyph_idx > 0 => {
-                self.buffer.with(|buff| {
+                self.buffer.with_untracked(|buff| {
                     let mut prev_word_idx = 0;
                     for (idx, _) in buff.unicode_word_indices() {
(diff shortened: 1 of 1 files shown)


dzhou121:
Thanks for working on it. But there's plan to replace the current `text_input` with a new implementation backed by 'text_editor`.

sgued:
Having only one text editor view seems like a great idea. Should this be merged nonetheless to fix programs using main?

dzhou121:
Makes sense, since we don't know when the new text input will be implemented :)

PR #556, opened 2024-08-21
Fill height

tamewild:
Setting in the editor that determines whether to fill the parent container's height or not. 

If disabled it allows shrinking to the height of its contents instead of filling.

Not sure if this is the right way to go about things.


tamewild:
this isnt right i think, mb.

PR #559, opened 2024-08-23
add get/update from fn

jrmoulton:
- Creates two new traits `SignalRead` and `SignalWrite` that just split out the accessing of the Rc data from the get and update traits. 

- Adds a new signal type that lets you define a base signal and then a custom getter function and a custom setter function. This type implements the new signal traits

- Changes the checkbox widget to just accept in input that implements the signal traits. 

- Changes the slider to use the more typical rusty `new` function instead of a `create_*`. (I'm leaning towards switching all of our APIs to this format). 

- Adds a few different `Slider::new_*` functions that accept inputs that implement different sets of the Signal traits. 

- Changes the slider example in the widget gallery to have the only state the the string for the text input and then the slider is driven by a `GetUpdateFn` with custom accessors. 

I don't love the name `GetUpdateFn` but I haven't come up with anything better


src/views/checkbox.rs
```diff
@@ -1,58 +1,42 @@
...
-        checkbox_svg(inbound_signal.read_only())
-            .keyboard_navigatable()
-            .on_click_stop(move |_| {
-                let checked = inbound_signal.get_untracked();
-                outbound_signal.set(!checked);
-            }),
-        move || outbound_signal.get(),
-    )
+pub fn checkbox(
+    checked: impl SignalGet<bool> + SignalUpdate<bool> + 'static + Copy,
```
dzhou121, line 21: I think a closure is still better here. It doesn't need to create a signal if you just pass a value in, e.g.

`checkbox(|| true)`

vs

`checkbox(create_rw_signal(true))`

dzhou121:
For `create_*`, I personally prefer this over "RwSignal::new()`, because it blends well with all the view creation functions. 

Also, it's simply easier to type in, I normally just put crrw and the auto completion will pick it up :)

jrmoulton:
Part of the reason that I like the `::new_*()` methods is because that works better for when there are multiple ways to constructor methods create a view. Like how I currently have it with slider. But I agree that in the simple case the` create_*` methods are cleaner. I'll just make sure to have a `create_*` function that creates the default view

PR #562, opened 2024-08-28
rename SignalWith track method

jrmoulton:
This avoids naming conflicts with the SignalRead track method


dzhou121:
Would it better to only keep one of them?

dzhou121:
Or give track its own trait SignalTrack?

jrmoulton:
Hmm yeah it's own trait would probably be good

PR #563, opened 2024-08-29
Keyframe Animations

jrmoulton:


jrmoulton:
This still needs to figure out how to control an animation. Start/stop/pause/resume

jrmoulton:
There are still some open questions in this API. 

One of the biggest things is that you can't animate between pixel values and percent. I have some ideas on how to solve that but each has problems. 

Another is a question of how should stacked animations interact. Currently they don't work well together if multiple animations have the same props. But I'm not sure if the correct behavior would be to interpolate between stacked animations?

Future performance optimizations:
- Add a cache in the animation to look up the last keyframe that had a particular prop. I have a comment with some scaffolding in the Animation struct

- When updating a keyframe, send a reactive update the same way we are with the state command instead of rebuilding the entire animation

jrmoulton:
It would still be good to make it so that the animations can resolve nested maps and to be able to interploate between px/pct/auto but I think this is good to go for now

PR #573, opened 2024-09-12
signal impl

jrmoulton:
Adds a `RefCell` to event handlers for the closures to borrow/copy the data mutably


jrmoulton:
Other Fn traits will probably need to be changed to FnMut but I've only changed these ones for now. Doing so caused no issues with the existing examples

PR #584, opened 2024-09-20
switch svg and button default to static

jrmoulton:


jrmoulton:
I think zoxc mentioned this a while ago and now I pretty strongly agree with it

dzhou121:
would something like this would be better

```rust
pub fn svg(svg_str: impl Into<String> + 'static) -> Svg {
    let id = ViewId::new();
    id.update_state(svg_str);
    Svg {
        id,
        svg_tree: None,
        svg_hash: None,
    }
    .class(SvgClass)
}

impl Svg {
    pub fn update(self, svg_str: impl Fn() -> String + 'static) -> Self {
        let id = self.id;
        create_effect(move |_| {
            let new_svg_str = svg_str();
            id.update_state(new_svg_str);
        });
        self
    }
}
```

so that we don't have `svg` `dyn_svg`

jrmoulton:
ah yeah that seems good. I think update makes sense.

jrmoulton:
used update_svg as the method to avoid confusing with View update method

dzhou121:
maybe consider using `value` so that we can use it for all views that needed it, e.g. `label` etc

dzhou121:
or `update_value`?

jrmoulton:
👍

PR #590, opened 2024-09-24
fix scale, rename transform to translate

jrmoulton:
@timsueberkrueb

### DIFF
--- a/vger/src/lib.rs
+++ b/vger/src/lib.rs
@@ -9,6 +9,7 @@ use floem_renderer::text::{CacheKey, TextLayout};
 use floem_renderer::{tiny_skia, Img, Renderer};
 use floem_vger_rs::{Image, PaintIndex, PixelFormat, Vger};
 use image::{DynamicImage, EncodableLayout, RgbaImage};
+use peniko::kurbo::Size;
 use peniko::{
     kurbo::{Affine, Point, Rect, Shape},
     BrushRef, Color, GradientKind,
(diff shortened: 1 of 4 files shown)


jrmoulton:
This doesn't claim to implement rotation. rotation would have to be even more special handling in the renderer

jrmoulton:
This also doesn't fix event handling. I'll do that in another PR

timsueberkrueb:
This is great, thank you @jrmoulton! It fixes font rendering when scaling.

> This doesn't claim to implement rotation. rotation would have to be even more special handling in the renderer

Maybe this gets easier if all transformations are consistently handled as a single affine transformation matrix (which is already partially the case). At the moment, with the vger renderer, we extract the individual properties (like translation and scale) from the transformation matrix instead of applying it directly. Maybe it makes sense to change this when we switch to Vello? I've seen that its API has first-class support for passing transformation matrices.

Anyways, I don't need rotation for my use case right now, but I think it would be nice to have eventually (for animations, spinners, etc).

vger/src/lib.rs
```diff
@@ -418,26 +419,33 @@ impl Renderer for VgerRenderer {
         let transformed_y = transform[1] * pos.x + transform[3] * pos.y + transform[5];
         let pos = Point::new(transformed_x, transformed_y);
 
-        let coeffs = self.transform.as_coeffs();
-        let scale = (coeffs[0] + coeffs[3]) / 2. * self.scale;
+        let scale_x = transform[0];
+        let scale_y = transform[3];
+
+        let scale = (transform[0] + transform[3]) / 2. * self.scale;
+        if scale.abs() < 0.1 {
```
timsueberkrueb, line 426: I think this is fine. One thing that could be considered in the future is rendering rectangles in place of the text with the appropriate text color and spacing when the overall font size is less than a certain threshold.

jrmoulton: Ah that would be good. I'll add a comment/todo for now

jrmoulton:
This can sometimes trigger some strange effects when animating scale. 

It doesn't happen consistently.

I've watched all of the scale values that affect the different properties while animating and they look correct. I now believe this is some bug in vger

https://github.com/user-attachments/assets/55eeda36-5079-49bb-8d15-4a778a939acc

jrmoulton:
> . Maybe it makes sense to change this when we switch to Vello? I've seen that its API has first-class support for passing transformation matrices.

yeah in my branch for vello this should already be handled.

jrmoulton:
https://github.com/user-attachments/assets/9d67747c-ec4c-4390-8bab-2d924189bee5

PR #592, opened 2024-09-26
Allow `tab` to have no active tabs.

hydra:
This is a common scenario when using a tabbed interface with one-tab per document without any documents having yet been loaded.

A running example using this can be found here:

https://github.com/hydra/floem/tree/tabbed-document-ui

I'm new to Floem, likely I made mistakes, so review this PR carefully.


dzhou121:
Do you mean for cases that tab children is empty?

hydra:
> Do you mean for cases that tab children is empty?

yes,  I note that in the `tab` implementation, there is a default value of `0` for the index for the active tab, as opposed to `None`.  See this change:

https://github.com/lapce/floem/pull/592/files#diff-7baac154f944b70ba151fa8121dc250311a7d472e5f14a485c48a02e749a6dd3L88

It's possible I'm missing something, as the API to the `tab()` isn't documented and it was very unclear to me what the purpose of 3rd argument (key_fn) was supposed to do, and wasn't obvious when looking at the layout example which is the only example that calls the `tab` method and always has a non-zero amount of tabs.

PR #605, opened 2024-09-30
Image api improvements

hydra:
While working with the `img` API I was frustrated to find that I had to create a vector of u8's even though I already had either a DynamicImage instance or PathBuf instance.  e.g. when creating a new file (`DynamicImage`) or opening a file (`PathBuf`).

I also noted that `img_dynamic` visibility is `pub(crate)` which meant I could not use that either.

This PR adds three well-named methods (commit 1) and deprecates the old one (commit 2).

See commits/changes.

### DIFF
--- a/src/views/img.rs
+++ b/src/views/img.rs
@@ -90,6 +90,7 @@ pub struct Img {
     content_node: Option<NodeId>,
 }
 
+#[deprecated]
 pub fn img(image: impl Fn() -> Vec<u8> + 'static) -> Img {
     img_from_bytes(image)
 }
--- a/src/views/img.rs
+++ b/src/views/img.rs
@@ -1,3 +1,4 @@
+use std::path::PathBuf;
 use std::rc::Rc;
 
 use floem_reactive::create_effect;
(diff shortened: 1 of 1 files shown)


charlescgs:
Hello, any particular reason this PR cannot be merged?

jrmoulton:
@charlescgs 

> Hello, any particular reason this PR cannot be merged?

It hasn't passed CI and typically won't be reviewed until it does. 

I do think though that we don't want to deprecate the `img` function. I think we can leave it as it is and just add the two new functions. Final approval is from dzhou121 though.

panekj:
> I do think though that we don't want to deprecate the `img` function

I agree, `img` is nice and convenient shorthand

charlescgs:
@panekj then how about this:
```rust
pub fn img(image: impl Fn() -> Vec<u8> + 'static) -> Img {
    let image = image::load_from_memory(&image()).ok();
    let width = image.as_ref().map_or(0, |img| img.width());
    let height = image.as_ref().map_or(0, |img| img.height());
    let data = Arc::new(image.map_or(Default::default(), |img| img.into_rgba8().into_vec()));
    let blob = Blob::new(data);
    let image = peniko::Image::new(blob, peniko::Format::Rgba8, width, height);
    img_dynamic(move || image.clone())
}

pub fn img_from_path(image: impl Fn() -> PathBuf + 'static) -> Img {
    let image = image::open(&image()).ok();
    let width = image.as_ref().map_or(0, |img| img.width());
    let height = image.as_ref().map_or(0, |img| img.height());
    let data = Arc::new(image.map_or(Default::default(), |img| img.into_rgba8().into_vec()));
    let blob = Blob::new(data);
    let image = peniko::Image::new(blob, peniko::Format::Rgba8, width, height);
    img_dynamic(move || image.clone())
}

pub fn img_from_image(image: impl Fn() -> peniko::Image + 'static) -> Img {
    img_dynamic(move || image())
}
```

PR #609, opened 2024-10-01
Transform view

timsueberkrueb:
Useful component for scroll views/pan zoom areas and similar applications. Not sure if we want this in floem, but I'm opening this for discussion.

### DIFF
--- a/examples/transform/Cargo.toml
+++ b/examples/transform/Cargo.toml
@@ -1,0 +1,8 @@
+[package]
+name = "transform"
+edition = "2021"
+license.workspace = true
+version.workspace = true
+
+[dependencies]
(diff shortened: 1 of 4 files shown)


jrmoulton:
What is the use case for this view if you can set the translation and scale on a view using the styles?

timsueberkrueb:
It doesn't go beyond being a convenience component for scroll/pan zoom areas. I'm also leaning towards not adding it to floem. Perhaps I can change the example to set the style directly.

timsueberkrueb:
Or add the transform view as part of the example instead?

timsueberkrueb:
I mostly care about having a test case/example for this application.

PR #618, opened 2024-10-12
Simplify dropdown, add constructors, add docs

jrmoulton:
image

This also removes the `dropdown` constructor function because it hides the fact that there are multiple constructors that you might want to use in different situations.


jrmoulton:
any comments @dzhou121 ?

jrmoulton:
image

PR #620, opened 2024-10-12
Views module docs

jrmoulton:


jrmoulton:
I'm experimenting with a few things here. 

I've added a prelude and I think it is a really nice change. Whenever I create a new file I always have to do `use module::*` for several modules anyways so I figured it was about time to add a prelude. 

I'm also experimenting with, in the views module, only doing `pub use module::*` for the most important views that are always used such as stacks, svg, container, text. I was already doing this for a few of the widgets that I had created such as dropdown and now I'm moving a few more to be like that. I feel like this is a good change. Leaving them in the modules makes the documentation for those items  and the whole views module more consistent.

I'm considering not doing `pub use module::*` at all inside of the views module and instead selectively publicly exporting them from the prelude. This would really help to clean up the views module documentation

PR #621, opened 2024-10-12
document dropdown menu

MightyBurger:
Cooked up a little something.

You should probably scrutinize this for technical accuracy. No hard feelings if you completely rewrite or scrap this.

### DIFF
--- a/src/views/dropdown.rs
+++ b/src/views/dropdown.rs
@@ -1,3 +1,4 @@
+//! Items related to the Dropdown view. See [dropdown].
 use std::{any::Any, rc::Rc};
 
 use floem_reactive::{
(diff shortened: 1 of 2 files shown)


MightyBurger:
Oops.... I did not see your pull request! Looks like my work might've been for naught. Ah well, I learned how rustdoc worked. Feel free to close this PR unless you want to use it.

jrmoulton:
Ah the example is good! Let's wait to see what happens/if there is any review on the PR I opened and then lets update this PR with those changes to add in the example. I agree that populating the list from an enum is a common use case and I think it would be good to have an example for it in the docs

jrmoulton:
@MightyBurger I've updated the code/docs for the dropdown in my PR. Could you update your changes to be up to date with the latest changes?

src/views/dropdown.rs
```diff
@@ -30,28 +31,120 @@ prop_extractor!(DropdownStyle {
...
-{
-    Dropdown::new(active_item, main_view, iterator, list_item_fn)
-}
-
-/// A dropdown widget
+/// Create a dropdown widget.
+///
+/// # Example
+///
+/// A common scenario is populating a dropdown menu using an enum.
```
jrmoulton, line 38: I think it would be good to have this in the docs for the Dropdown struct. The example in the widget gallery for the dropdown does exactly this so it might be good to have the docs also reference that example.

src/views/dropdown.rs
```diff
@@ -30,28 +31,120 @@ prop_extractor!(DropdownStyle {
...
+/// # }
+/// #
+/// # fn character_select() -> impl floem::IntoView {
+/// use floem::{reactive::*, views::*, IntoView};
+///
+/// let mode = RwSignal::new(Character::Ori);
+///
+/// dropdown::dropdown(
+///     move || mode.get(),
+///     move |m| text(m.to_string()).into_any(),
```
jrmoulton, line 97: the `.to_string()` isn't actually necessary. text and label can directly take an item that impls display

src/views/dropdown.rs
```diff
@@ -30,28 +31,120 @@ prop_extractor!(DropdownStyle {
...
+///     Character::iter(),
+///     move |m| text(m.to_string()).into_any(),
+/// )
+/// .on_accept(move |new| mode.set(new))
+/// # }
+/// ```
+///
+/// You can display the selected item differently than the list of items.
```
jrmoulton, line 105: I think just the single example showing how to build and use the enum iter is sufficient.

jrmoulton (changes requested):

PR #635, opened 2024-10-21
Switch to Vello (new PR)

jrmoulton:
Vello uses a lot more memory than vger does. 

In debug mode, the widget gallery with vger uses about 90MB of memory, vello uses over 400. (specific numbers depend greatly on window size). 

similar in release mode. 

Each new window will use another another renderer which will use another 400-600 MB of memory which gets excessive quickly. 

Need to fix:
 - I have a bug where text that is italic / bold is just rendered as normal. I think this is something I am doing wrong. 
 - Not all svgs that resvg rendered are getting rendered by vello_svg. I'm not sure why this is. If it is in vello_svg it is failing without warning.
 - svg rendering in general has issues (sizes, colors, unwanted background color)

Marking as draft because:
 - vello doesn't have the font_embolden. (Personally I don't like the look on macOS with font embolden so I am fine with that.)
 - high memory usage

 - Need to verify that vello actually performs comparable to vger
 
 I do really want the features (and correctness) vello offers. Even with the tradeoffs I'd much prefer to be working on floem with vello as the renderer. 
@DJMcNab


DJMcNab:
It's really great to see this.

Why would a second window require a second `Renderer`?

We have not put much effort into reducing memory usage yet - some work has started in https://github.com/linebender/vello/pull/606 but has stalled due to a lack of review bandwidth.

Italic/bold text should be handled by your text rendering library providing you with the bold/italic font for that run, or else providing the correct emulated variable font parameters.
vello_svg has not been carefully tested for correctness/completeness. Contributions in that direction would be welcome.

jrmoulton:
> Why would a second window require a second Renderer?

Sounds like it doesn't need to. I'll see about fixing that.

> We have not put much effort into reducing memory usage yet - some work has started in https://github.com/linebender/vello/pull/606 but has stalled due to a lack of review bandwidth.

👍 

> Italic/bold text should be handled by your text rendering library providing you with the bold/italic font for that run, or else providing the correct emulated variable font parameters.

Yeah I've probably just messed something up.

> vello_svg has not been carefully tested for correctness/completeness. Contributions in that direction would be welcome.

I'll see what I can do there.

dzhou121:
I think maybe we can still keep vger, and control via feature flags?

jrmoulton:
Yeah that would probably be the way to go for now. That would allow us to keep improving the vello integration without affecting lapce. 

I'll add back vger and then mark this as ready for review and then open issues for the current issues that I've mentioned

examples/animations/src/main.rs
```diff
@@ -37,7 +38,7 @@ fn app_view() -> impl IntoView {
             .animation(move |_| animation.get())
             .animation(move |a| {
                 a.keyframe(100, |kf| {
-                    kf.style(|s| s.border(5).border_color(Color::PURPLE))
+                    kf.style(|s| s.border(Stroke::new(5.)).border_color(Color::PURPLE))
```
dzhou121, line 41: Maybe we could make a trait to convert numbers to `StrokeWrap` so that we can still write it the old way.

jrmoulton: Ah we can. I first started converting them to Stroke::new and after it was too much of a pain I added impls for From<num> for StrokeWrap. I just haven't changed these examples back to just numbers but it would work if I did.

jrmoulton: @dzhou121 do you want these changed back?

dzhou121: Yes please

PR #646, opened 2024-10-27
add: disable default view events

jrmoulton:
not a breaking change. unless code adds id.disable_default_event, no behavior of event handling will change

### DIFF
--- a/examples/animations/src/main.rs
+++ b/examples/animations/src/main.rs
@@ -1,7 +1,6 @@
 use floem::{
     animate::Animation,
     event::EventListener as EL,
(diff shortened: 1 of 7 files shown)


jrmoulton:
any opposition to adding this?

dzhou121:
What’s the use case for this?

jrmoulton:
I have a text_input that I want to not process any events, but I don't want to mark it as disabled because that will stop it from being able to be focused using the keyboard and tab.

jrmoulton:
https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault

PR #648, opened 2024-10-28
dont send pointer events to overlap siblings

jrmoulton:


dzhou121:
Do you know if web allow it?

jrmoulton:
yeah web by default will block pointer events from overlapping sibling elements. css does have a pointer-events property that will stop the element from receiving any pointer events at all
https://drafts.csswg.org/css-ui/#hit-testing

jrmoulton:
I could make this configurable on an application/window level so that applications that don't want to check this during layout can avoid it

jrmoulton:
https://claude.site/artifacts/33e3ae52-b07c-4a85-93ec-3e5b8d687165

example generated with claude

dzhou121:
Got you now. 

Maybe we could make the implementation easier. In the default view event handler, if the view doesn't have any children or no children has handled the pointer event, it stops pointer event propagation in this view.

jrmoulton:
Ah yeah that could work. I think the logic would be if the pointer event was contained in a child, and it didn't handle it, then don't send it to any of the next children. That would be a lot simpler

jrmoulton:
Hahah yeah that's a lot simpler

jrmoulton:
Hmmm.. actually it might not be working as I expected

PR #662, opened 2024-11-04
fix view transition

jrmoulton:
https://github.com/user-attachments/assets/1bae04a4-dcaf-4ace-9732-ab947f7c3946

this also fixes an issue in my implementation of the vello renderer where I wasn't properly accessing the correct font index


jrmoulton:
@timsueberkrueb as an FYI, the transform effects are much more robust/correct/complete when using vello as the renderer (although I still haven't fixed event handling)

timsueberkrueb:
@jrmoulton That's awesome, thank you!

PR #666, opened 2024-11-05
add "animation", "draggable", "dropped file" example to widget-gallery  @tk103331

tk103331:
image

image

image


jrmoulton (changes requested):
I think this is a good change. I think we should go ahead and remove the stand alone examples as well so they aren't repeated.

PR #667, opened 2024-11-05
event pass through

### LINKED ISSUE #665: panic opening inspector when using tiny skia
probably broke when I changed the renderer to output a peniko image instead of the dyn imaeg

jrmoulton:
This makes it so that you can configure events to be allowed to be passed to overlapping siblings. This is necessary for the widget gallery that creates highlight overlays over items. 

This also fixes #665

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -452,6 +452,10 @@ impl<'a> EventCx<'a> {
             return true;
         };
 
+        if !id.state().borrow().pointer_events {
+            return false;
+        }
+
         let layout_rect = id.layout_rect();
         let Some(layout) = id.get_layout() else {
             return false;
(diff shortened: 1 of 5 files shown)


dzhou121:
Do we want to do it via something similar to `pointer-events: none;`?

jrmoulton:
What about an enum: Block, Passthrough, None?

jrmoulton:
I also wasn't prefectly sure on the naming. pointer-events:none is what it is on the web so it might just be good to use that since it will be common/discoverable as the name

dzhou121:
Yeah following with the web's name would be fine since we already have done lots of that.

jrmoulton:
Hmm. I'll skip on the enum for now. just the two states covers all current use cases so I'll just do that

PR #672, opened 2024-11-07
fix scroll bar

### LINKED ISSUE #671: Scroll issue
Hey, I encountered an issue, when having scroll view, I am unable to scroll with mouse and handle. Only scroll wheel works.

It doesn't work as well in the 'toto-complex' example.

I am using Windows 11 with newest floem master on vello.

jrmoulton:
fixes #671 

Fixes the bounds check and in doing so now makes the assumption that the scroll bars should always be in view regardless of the distance scrolled. ie if there is horizontal scrolling the vertical bar should remain fixed in place. 

this is the behavior I would expect so I'm not sure why it wasn't like that before

### DIFF
--- a/src/views/scroll.rs
+++ b/src/views/scroll.rs
@@ -449,8 +449,8 @@ impl Scroll {
 
             if let Some(color) = track_style.color() {
                 let mut bounds = bounds - scroll_offset;
-                bounds.y0 = self.content_rect.y0;
-                bounds.y1 = self.content_rect.y1;
+                bounds.y0 = self.total_rect.y0;
+                bounds.y1 = self.total_rect.y1;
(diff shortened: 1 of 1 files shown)


src/views/scroll.rs
```diff
@@ -513,10 +513,10 @@ impl Scroll {
         let top_y_offset = ((self.total_rect.height() - length) * percent_scrolled).ceil();
         let bottom_y_offset = top_y_offset + length;
 
-        let x0 = scroll_offset.x + self.total_rect.width() - bar_width - bar_pad;
```
jrmoulton, line 516: I don't know why this was here before

jrmoulton:
Ok never mind. this has issues

PR #673, opened 2024-11-07
remove redundant examples

jrmoulton:


dzhou121:
I feel like we should keep this. 

One example for one feature would be much easier for user to look at than the widget gallery, where they need to find where certain things are.

jrmoulton:
I feel like that is generally true, but I felt like these were a good fit for the widget gallery (with maybe the exception of animation) because they were demonstrating how to build widget-like functionality even though it isn't a built in widget.

jrmoulton:
I could still add them back though and we'll just have to manage the duplicated code

dzhou121:
> I could still add them back though and we'll just have to manage the duplicated code

Maybe we could let widget gallery to import the individual ones?

PR #674, opened 2024-11-07
fix drag jitter

jrmoulton:
this fixes some jitter when dragging views. because the view could have moved between the time that it is painted, the computed offset can be outdated. This causes lots of jitter in the draggable example. 

I feel like there should be a simpler way of doing this but I couldn't figure out another way


dzhou121:
Do you mean the pointer could have moved between the time that it is painted? But if we don't receive the pointermove event, there's no information the pointer has been moved.

Do I understand this issue correctly?

dzhou121:
Thanks for the clarification. 

Because the dragged view should be always follows the pointer, maybe we can simply it by storing the offset of the view relative to the pointer at drag start, and then at paint time, we know where it should be according to the pointer position. E.g. we don't use the relative value with the original view which could have moved as well, but use the relative value to the pointer and it should always be right.

jrmoulton:
https://github.com/user-attachments/assets/77b7fd25-190e-4cd5-9bc3-c6360da6ed56

PR #676, opened 2024-11-08
add files example to widget-gallery && add "Open In Window" in widget-gallery

tk103331:
add files example to widget-gallery
image

add "Open In Window" in widget-gallery
image

remove repeated examples: animation,draggable,dropped file,files


dzhou121:
I still personally prefer to have standalone examples for individual features. For example, when I try to write something in Axum, and it's much easier to go to the individual examples I want to achieve because it's much easier to have only a main.rs.

dzhou121:
@tk103331 could you still keep the "files" example, and in widget-gallery you can use "files" as a crate?

PR #677, opened 2024-11-08
use the "files" example as a crate,  and add the "files" example to widget-gallery

tk103331:
- use the "files" example as a crate

- add files example to widget-gallery

image

- add "Open In Window" in widget-gallery

image


dzhou121:
Thanks!

PR #683, opened 2024-11-13
simplify widget gallery by using list

jrmoulton:
image


charlescgs:
Thanks for that, I was about to make an issue for that haha :D
Even recorded an example:

https://github.com/user-attachments/assets/0da68511-5cbc-4945-b5cd-dbf0419c11a3

PR #691, opened 2024-11-15
make signals Not Send

### LINKED ISSUE #534: [docs] suggestion: explicity remark that signals cannot be used from other threads
Using signals from another thread (e.g. `set()`) leads to the following error at runtime:
```
thread '<unnamed>' panicked at /home/florian/.cargo/registry/src/index.crates.io-6f17d22bba15001f/floem_reactive-0.1.0/src/signal.rs:274:39:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I was able to resolve this using one of the methods from `floem::ext_event`.
But that was only the case because:
a) I already suspected using signals across multiple threads might not work
b) the combination of "signal + thread" in GH's issue search lead to a PR mentioning the aforementioned `create_signal_from_channel`/`update_signal_from_channel` functions.

P.S.: `floem` has been a pleasure to use so far :)

jrmoulton:
Fixes #534

This makes all signals !Send. 

```rust
fn counter_view() -> impl IntoView {
    let mut counter = RwSignal::new(0);

    std::thread::spawn(move || {
        counter.get();
    });

    h_stack((
        button("Increment").action(move || counter += 1),
        label(move || format!("Value: {counter}")),
        button("Decrement").action(move || counter -= 1),
    ))
    .style(|s| s.size_full().items_center().justify_center().gap(10))
}
```

Gives this error message

```
 1  error[E0277]: `*const ()` cannot be sent between threads safely                                                                                     ▐
    --> examples/counter-simple/src/main.rs:10:24                                                                                                       ▐
     |                                                                                                                                                  ▐
 10  |       std::thread::spawn(move || {                                                                                                               ▐
     |       ------------------ ^------                                                                                                                 ▐
     |       |                  |                                                                                                                       ▐
     |  _____|__________________within this `{closure@examples/counter-simple/src/main.rs:10:24: 10:31}`                                                ▐
     | |     |                                                                                                                                          ▐
     | |     required by a bound introduced by this call                                                                                                ▐
 11  | |         counter.get();                                                                                                                         ▐
 12  | |     });                                                                                                                                        ▐
     | |_____^ `*const ()` cannot be sent between threads safely                                                                                        ▐
     |                                                                                                                                                  ▐
     = help: within `{closure@examples/counter-simple/src/main.rs:10:24: 10:31}`, the trait `std::marker::Send` is not implemented for `*const ()`, whic▐
 note: required because it appears within the type `floem::floem_reactive::id::NotThreadSafe`                                                           ▐
    --> /Users/jaredmoulton/Developer/floem/reactive/src/id.rs:7:8                                                                                      ▐
     |                                                                                                                                                  ▐
 7   | struct NotThreadSafe(*const ());                                                                                                                 ▐
     |        ^^^^^^^^^^^^^                         
 ```
     
ExtActions still work because a new signal is introduced that is just a copy of Trigger but that has an unsafe impl of Send and Sync

### DIFF
--- a/examples/counter-simple/src/main.rs
+++ b/examples/counter-simple/src/main.rs
@@ -1,26 +1,16 @@
-use floem::{
-    reactive::create_signal,
-    views::{button, label, Decorators},
-    IntoView,
-};
-
-fn app_view() -> impl IntoView {
-    // Create a reactive signal with a counter value, defaulting to 0
-    let (counter, mut set_counter) = create_signal(0);
-
-    // Create a vertical layout
-    (
-        // The counter value updates automatically, thanks to reactivity
-        label(move || format!("Value: {counter}")),
-        // Create a horizontal layout
-        (
-            button("Increment").action(move || set_counter += 1),
-            button("Decrement").action(move || set_counter -= 1),
-        ),
-    )
-        .style(|s| s.flex_col())
-}
+use floem::prelude::*;
 
 fn main() {
-    floem::launch(app_view);
+    floem::launch(counter_view);
+}
+
+fn counter_view() -> impl IntoView {
+    let mut counter = RwSignal::new(0);
+
+    h_stack((
+        button("Increment").action(move || counter += 1),
+        label(move || format!("Value: {counter}")),
+        button("Decrement").action(move || counter -= 1),
+    ))
+    .style(|s| s.size_full().items_center().justify_center().gap(10))
 }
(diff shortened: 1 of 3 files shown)


dzhou121:
I actually was thinking of make signal to be able to send + sync

jrmoulton:
> I actually was thinking of make signal to be able to send + sync

I would love that. I wish I would've made this change before the release but I figure this is still a good change for now. 

What do you have in mind for making them send+sync?

dzhou121:
I started to think about when I'm doing the winit PR, and feel like 80% will be done by simply change the Rc<RefCell<>> to Arc<RwLock<>>. And the rest would be fixing bugs.

jrmoulton:
I'll merge this for now but I do think using arc/RwLock would be a good change

PR #698, opened 2024-11-22
Add tests for checkbox

sundaram123krishnan:


jrmoulton:
These tests aren't really testing the checkbox. They are just testing that signals get updates. 

I think for widgets the important api to test is the update method from the view trait and testing that directly. 

Because the checkbox doesn't directly use the update method I don't think it needs any tests. 

I'll close this for now but we are definitely interested in having tests but proper testing isn't trivial yet. Some work probably needs to be done to ensure a proper test environment that works with window events and changes that require multiples passes

PR #701, opened 2024-11-27
fix draggable

### LINKED ISSUE #700: Draggable sidebar example not working
Hi,

Has [this example](https://github.com/lapce/floem/blob/main/examples/layout/src/draggable_sidebar.rs) regressed? When I hover I see no border to drag:

https://github.com/user-attachments/assets/e9cb94c0-fe7f-454a-bf98-0b22465ef166

If I change this to red `.border_color(Color::rgb8(255, 0, 0))` I can see it, but can't drag it:

```rust
let dragger = label(|| "")
        .style(move |s| {
            s.position(Position::Absolute)
                .z_index(10)
                .inset_top(0)
                .inset_bottom(0)
                .inset_left(sidebar_width.get())
                .width(10)
                .border_left(1)
                .border_color(Color::rgb8(255, 0, 0))
                .hover(|s| {
                    s.border_left(2)
                        .border_color(Color::rgb8(41, 98, 218))
                        .cursor(CursorStyle::ColResize)
                })
                .apply_if(is_sidebar_dragging.get(), |s| {
                    s.border_left(2).border_color(Color::rgb8(41, 98, 218))
                })
        })
        .draggable()
        .dragging_style(|s| s.border_color(Color::RED))
```

Screenshot 2024-11-24 at 17 34 22

jrmoulton:
Fixes #700


jrmoulton:
This was a regression. This broke because we made it so that if there are overlapping elements, if the element on top doesn't handle the event, the event isn't passed to the next item. A simple fix is just moving the dragger to be the last item in the stack so that it is considered "on top".

PR #702, opened 2024-11-29
Delay window visibility to prevent white background on app launch

### LINKED ISSUE #468: Why does a white background appear when Lapce or Floem Application  is launched, and how can it be eliminated?
image
**As shown in the above image, there is a white background that appears upon startup; it's very unsightly and affects the visual experience. It is recommended to eliminate it.**

Long0x0:
This PR delays window visibility until renderer is initialized to prevent the white background flash on app launch.
Tested on Windows 11 and Ubuntu.
Fix #468.


jrmoulton:
Thanks for this fix. I think this is the reasonable default

PR #707, opened 2024-12-08
Expose TouchpadMagnify event

andi23rosca:
First PR contributing here :D

Noticed floem is not exposing the winit TouchpadMagnify event to the users

### DIFF
--- a/src/app_handle.rs
+++ b/src/app_handle.rs
@@ -234,7 +234,9 @@ impl ApplicationHandle {
             WindowEvent::MouseInput { state, button, .. } => {
                 window_handle.mouse_input(button, state);
             }
-            WindowEvent::TouchpadMagnify { .. } => {}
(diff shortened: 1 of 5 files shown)


src/app_handle.rs
```diff
@@ -234,7 +234,9 @@ impl ApplicationHandle {
             WindowEvent::MouseInput { state, button, .. } => {
                 window_handle.mouse_input(button, state);
             }
-            WindowEvent::TouchpadMagnify { .. } => {}
+            WindowEvent::TouchpadMagnify { delta, .. } => {
```
jrmoulton, line 237: This should also expose the phase

jrmoulton (changes requested):
This looks good but I think it would be good to expose the phase as well

src/app_handle.rs
andi23rosca: done ✅

jrmoulton:
Thanks!

PR #712, opened 2024-12-24
1. support rfd synchronous api

jm-observer:


panekj:
Why add this? rfd doesn't have sync API in `xdg-portal`

PR #713, opened 2024-12-24
1. support floem-editor-core serde feature

jm-observer:


Cargo.toml
```diff
@@ -101,6 +101,7 @@ serde = [
   "peniko/serde",
   "lapce-xi-rope/serde",
   "smallvec/serde",
+  "floem-editor-core/serde"
```
panekj, line 104:
```suggestion
  "floem-editor-core?/serde"
```

panekj (changes requested):

PR #716, opened 2024-12-29
impl SignalWrite for RwSignal

macmv:
impl's `SignalWrite` for `RwSignal`.

I find myself writing code like this every so often:
```rs
let signal = RwSignal::new(/* ... */);

// later...
signal.write_only().write().borrow_mut().my_mutate_fn(/* ... */);
```

And, it would be nice to not need to convert my signal to writeonly first:
```rs
let signal = RwSignal::new(/* ... */);

// with this change:
signal.write().borrow_mut().my_mutate_fn(/* ... */);
```

### DIFF
--- a/reactive/src/signal.rs
+++ b/reactive/src/signal.rs
@@ -273,6 +273,12 @@ impl<T> SignalUpdate<T> for RwSignal<T> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


jrmoulton:
This is great. Thanks!

PR #723, opened 2025-01-03
The editor automatically completes '`' and '<'

jm-observer:


dzhou121:
I don't think '<' should be auto completed. e.g. `if i < 10`

PR #724, opened 2025-01-03
1. Combined with the inspector, this feature can help users identify …

jm-observer:
1. Combined with the inspector, this feature can help users identify which view handles or stops the event. It is very useful in large projects.
2. maybe remove log crate


jm-observer:
I am not familiar with the tracing crate, so maybe someone else can do it if needed.

PR #727, opened 2025-01-06
Enhance `text_input` component.

anishsinha-io:
The placeholder inside of a text input will now, by default, inherit the font family of it's parent text input. However, it may also be overridden. The `input` example inside the widget gallery was modified to demonstrate some of the new behavior.


dzhou121:
Thanks!

PR #733, opened 2025-01-12
Fix `create_signal_from_tokio_channel` feature

anishsinha-io:
There was an apparent type error where a `Trigger` was being passed to `crate::ext_event::register_ext_trigger; instead of an `ExtSendTrigger`. It was under a feature flag and isn't used in any examples which is probably why it evaded detection.

### DIFF
--- a/src/ext_event.rs
+++ b/src/ext_event.rs
@@ -44,9 +44,19 @@ impl ExtSendTrigger {
     }
 }
 
+pub trait ExtEvent {
+    fn create_ext_send_trigger(self) -> ExtSendTrigger;
+}
+
 pub fn create_trigger() -> ExtSendTrigger {
     ExtSendTrigger {
         signal: create_rw_signal(()),
+    }
+}
+
+impl ExtEvent for Scope {
+    fn create_ext_send_trigger(self) -> ExtSendTrigger {
+        with_scope(self, create_trigger)
     }
 }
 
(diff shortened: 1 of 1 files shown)


anishsinha-io:
I also tested this fix in a nontrivial project which uses `create_signal_from_tokio_channel` so at least I can confirm that the behavior of the patched function is as I expected.

src/ext_event.rs
```diff
@@ -44,12 +44,22 @@ impl ExtSendTrigger {
     }
 }
 
+pub trait ExtEvent {
+    fn create_ext_send_trigger(self) -> ExtSendTrigger;
+}
+
```
jrmoulton, line 50: Why create this trait?

jrmoulton:
This looks good (and yes was a bug) but I'm not sure I understand the reason for that trait

src/ext_event.rs
anishsinha-io: Without the trait, I can't do this:

```rust
let trigger = cx.create_ext_send_trigger();
```

because `create_ext_send_trigger` can't be implemented in the crate where `Scope` is defined since `ExtSendTrigger` is defined in `ext_event.rs` outside of the reactive crate.

src/ext_event.rs
```diff
@@ -202,7 +212,7 @@ pub fn create_signal_from_tokio_channel<T: Send + 'static>(
     mut rx: tokio::sync::mpsc::UnboundedReceiver<T>,
 ) -> ReadSignal<Option<T>> {
     let cx = Scope::new();
-    let trigger = cx.create_trigger();
+    let trigger = cx.create_ext_send_trigger();
```
jrmoulton, line 215:
```suggestion
    let trigger = with_scope(cx, ExtSendTrigger::new);
```

jrmoulton:
Thanks!

PR #740, opened 2025-01-21
Update to resvg 0.44

waywardmonkeys:


waywardmonkeys:
when an update to vello 0.4 happens, it is easier if these are in sync since `vello_svg` which supports Vello 0.4 is on the 0.44 release of `usvg`.

PR #741, opened 2025-01-22
Update wgpu, Peniko, and Vello

waywardmonkeys:
Update to Vello 0.4, Peniko 0.3.1, and wgpu 23.0.1. This brings with it the new `color` crate that Vello and Peniko use.

This currently disables interpolation of gradients.

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -3,6 +3,7 @@
 use floem_reactive::create_updater;
 use floem_renderer::text::{LineHeightValue, Weight};
 use im_rc::hashmap::Entry;
+use peniko::color::{palette, HueDirection};
 use peniko::kurbo::{Point, Stroke};
 use peniko::{Brush, Color, ColorStop, ColorStops, Gradient, GradientKind};
 use rustc_hash::FxHasher;
@@ -219,12 +220,12 @@ impl StylePropValue for Color {
                 .width(22.0)
                 .height(14.0)
                 .border(1.)
-                .border_color(Color::WHITE.multiply_alpha(0.5))
+                .border_color(palette::css::WHITE.with_alpha(0.5))
                 .border_radius(5.0)
         });
         let color = stack((color,)).style(|s| {
             s.border(1.)
-                .border_color(Color::BLACK.multiply_alpha(0.5))
+                .border_color(palette::css::BLACK.with_alpha(0.5))
                 .border_radius(5.0)
                 .margin_left(6.0)
         });
@@ -236,19 +237,7 @@ impl StylePropValue for Color {
     }
 
     fn interpolate(&self, other: &Self, value: f64) -> Option<Self> {
-        let r = f64::interpolate(&(self.r as f64), &(other.r as f64), value)
-            .unwrap()
-            .round() as u8;
-        let g = f64::interpolate(&(self.g as f64), &(other.g as f64), value)
-            .unwrap()
-            .round() as u8;
-        let b = f64::interpolate(&(self.b as f64), &(other.b as f64), value)
-            .unwrap()
-            .round() as u8;
-        let a = f64::interpolate(&(self.a as f64), &(other.a as f64), value)
-            .unwrap()
-            .round() as u8;
-        Some(Color { r, g, b, a })
+        Some(self.lerp(*other, value as f32, HueDirection::default()))
     }
 }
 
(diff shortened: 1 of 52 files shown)


waywardmonkeys:
This is a draft and not fully working yet (doesn't compile). The actual update of Vello isn't here yet until I finish up the changes for brush and gradient interpolation to at least get them to compile.

vger/Cargo.toml
```diff
@@ -16,5 +16,5 @@ wgpu = { workspace = true }
 futures = "0.3.26"
 anyhow = "1.0.69"
 swash = "0.1.8"
-floem-vger-rs = { version = "0.3.1", package = "floem-vger" }
+floem-vger-rs = { git = "https://github.com/lapce/vger-rs.git", rev = "b2441cc85b1d555a3c74e2cdced08382460759df", version = "0.3.1", package = "floem-vger" }
```
waywardmonkeys, line 19: This will need a release before this can land.

tiny_skia/src/lib.rs
```diff
@@ -51,8 +52,9 @@ pub struct TinySkiaRenderer<W> {
     cache_color: CacheColor,
 
     image_cache: HashMap<Vec<u8>, (CacheColor, Rc<Pixmap>)>,
+    // The `u32` is a color encoded as a u32 so that it is hashable and eq.
     #[allow(clippy::type_complexity)]
-    glyph_cache: HashMap<(CacheKey, Color), (CacheColor, Option<Rc<Glyph>>)>,
+    glyph_cache: HashMap<(CacheKey, u32), (CacheColor, Option<Rc<Glyph>>)>,
```
waywardmonkeys, line 57: With `color`, the `Color` (really `AlphaColor<Srgb>`) is using `f32`, so it isn't `Hash` or `Eq`. We have a way to work around this with a wrapper type, but since here we're just dropping down to u8 colors anyway, I thought this might be okay here. I'm not sure though.

src/style.rs
```diff
@@ -236,19 +237,7 @@ impl StylePropValue for Color {
...
-            .unwrap()
-            .round() as u8;
-        let b = f64::interpolate(&(self.b as f64), &(other.b as f64), value)
-            .unwrap()
-            .round() as u8;
-        let a = f64::interpolate(&(self.a as f64), &(other.a as f64), value)
-            .unwrap()
-            .round() as u8;
-        Some(Color { r, g, b, a })
+        Some(self.lerp(*other, value as f32, HueDirection::Increasing))
```
waywardmonkeys, line 240: Much shorter!

waywardmonkeys (commented):
The bulk of the changes are quite dull.

* Color constants are referenced from `color::palette::css` rather than named constants on the `Color` type.
* Use `with_alpha` rather than `multiply_alpha` since we're just using this to set the alpha.
* Use `Color::from_rgb8` and `Color::from_rgba8` rather than `Color::rgb8` and `Color::rgba8`.

jrmoulton (approved):
to this point all of these changes look good and I think it's fine to leave the gradient interpolation commented out

jrmoulton:
Thanks for doing this! It is very appreciated

PolyMeilex:
This somehow causes regression in inspector view rendering with Vulkan:

image

With OpenGL it seems to work

jrmoulton:
@PolyMeilex  It looks like that is rendering with tiny_skia and not any of the gpu renderers. If the change broke something it might be wgpu compatibility with your gpu.

PolyMeilex:
> @PolyMeilex  It looks like that is rendering with tiny_skia and not any of the gpu renderers. If the change broke something it might be wgpu compatibility with your gpu. 

Yeah, the vulkan one to be precise, that would explain why it works when I force opengl, I will debug this tomorrow to see what errors wgpu initialisation throws.

PolyMeilex:
Yep, new wgpu no longer sees my vulkan adapter
```
Adapter 0:
	         Backend: Vulkan
	            Name: llvmpipe (LLVM 19.1.7, 256 bits)
	        VendorID: 0x10005
	        DeviceID: 0x0
	            Type: Cpu
	          Driver: llvmpipe
	      DriverInfo: Mesa 25.0.0 (LLVM 19.1.7)
	WebGPU Compliant: true
Adapter 1:
	         Backend: Gl
	            Name: AMD Radeon Graphics (radeonsi, raphael_mendocino, LLVM 19.1.7, DRM 3.61, 6.13.5-200.fc41.x86_64)
	        VendorID: 0x1002
	        DeviceID: 0x0
	            Type: Other
	          Driver: <empty>
	      DriverInfo: 4.6 (Core Profile) Mesa 25.0.0
	WebGPU Compliant: false
```
So that's not a floem issue

PR #745, opened 2025-01-25
Export IconExtWindows

cyberdoors:
Export IconExtWindows, support loading icons from resources

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -6,6 +6,7 @@ pub use winit::window::Theme;
 pub use winit::window::WindowButtons;
 pub use winit::window::WindowId;
 pub use winit::window::WindowLevel;
+#[cfg(windows)]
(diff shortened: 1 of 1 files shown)


src/window.rs
```diff
@@ -6,6 +6,7 @@ pub use winit::window::Theme;
 pub use winit::window::WindowButtons;
 pub use winit::window::WindowId;
 pub use winit::window::WindowLevel;
+pub use winit::platform::windows::IconExtWindows;
```
panekj, line 9:
```suggestion
#[cfg(windows)]
pub use winit::platform::windows::IconExtWindows;
```

panekj (changes requested):

jrmoulton:
This looks good. It just needs rustfmt to run.

panekj:
You need to fix formatting issue with rustfmt too

PR #749, opened 2025-01-28
Update to objc2 0.6.0 family

waywardmonkeys:


waywardmonkeys:
`winit` and `wgpu` are still on the old versions, so not sure if you want to land this now vs later.

PR #752, opened 2025-02-03
make advanced layout possible

jrmoulton:
These items are necessary in order to do advanced layout. I'm marking as draft because I'm going to see if it's possible to just add more to the API to access just the necessary features


charlescgs:
What is the future of this one?

jrmoulton:
This could probably be merged. I don't think there is a good reason not to. I have continued to use it in my video editor

charlescgs:
Then let's do it:)

PR #754, opened 2025-02-05
Prioritize redraws over processing updates

jrmoulton:
Previously, if Floem couldn't process an entire update per frame, then no new frames would be rendered until it could catch up. This prioritizes redraws so that frame rates can scale down gracefully as required.


dzhou121:
It feels to me this approach will cause updates stuck until the next time events etc that triggers `process_update`.

PR #756, opened 2025-02-08
1. Add default styles to several views.

jm-observer:


jrmoulton:
It might be better to just have the PointerEventsProp not be wrapped in None. 

What behavior does this change? Why these views specifically?

jm-observer:
> It might be better to just have the PointerEventsProp not be wrapped in None.
> 
> What behavior does this change? Why these views specifically?

While troubleshooting the click event, I found that the is_pointer_none of these views is being calculated as true, but I’m not sure why.
```
#context.rs
        let is_pointer_none = event.is_pointer()
            && view_state.borrow().computed_style.get(PointerEventsProp)
                == Some(PointerEvents::None);
```

jrmoulton:
Floem doesn't default any views to use Some(PointerEvents::None). Doing a symbol search confirms that so, if that is true, I don't think these changes belong in Floem.

jm-observer:
> Floem doesn't default any views to use Some(PointerEvents::None). Doing a symbol search confirms that so, if that is true, I don't think these changes belong in Floem.

try my project, : https://github.com/jm-observer/cozy-lapce

PR #761, opened 2025-02-09
1. `PointerEventConsumed::Yes` will cause sibling (child) views to be…

jm-observer:
`PointerEventConsumed::Yes` will cause sibling (child) views to be unable to receive mouse events.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -447,7 +447,7 @@ impl EventCx<'_> {
             }
         }
 
-        (EventPropagation::Continue, PointerEventConsumed::Yes)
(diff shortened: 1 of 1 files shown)


dzhou121:
This is intentional. Set PointerEvent to None if you don't want the view to intercept mouse events.

jm-observer:
After upgrading Floem for Lapce, you should find it difficult to locate the views where None needs to be added.And if set to `PointerEventConsumed::No`, users can not modify anything

dzhou121:
You can take a look at https://github.com/lapce/lapce/pull/3622 to have some idea of the places that needs it.

PR #774, opened 2025-02-15
Add table view

jrmoulton:
draft because this is somewhat fragile. 

position in the virtualized table is handled internally by the table by manually setting a padding_top. So if the caller of a table ever manually sets a padding top, it will break the table. I did this because using a before_node like how the virtual list does breaks the grid layout. 

I've also had to hard code in the column headers by painting them in a position that is different from the layout. In order to get this to work without flashing the column headers I had to add a margin top internally to the table. 

These hacks wouldn't be necessary if we had layout support in taffy for subgrids and for position: sticky.


jrmoulton:
I think I want to rework this so that the table doesn't use taffy for layout

ansarizafar:
Is there any update. I need table view for my new project.

PR #775, opened 2025-02-16
Default to std::sync::mpsc, introduce `crossbeam` feature

charlescgs:
Functions `create_signal_from_channel` and `update_signal_from_channel` were using crossbeam version of the channels, what shouldn't be a default.

This PR introduce std channels as a default option and hides crossbeam counterparts behind the 'crossbeam' feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,7 +58,7 @@ tokio = { version = "1", features = ["sync", "rt"], optional = true }
 raw-window-handle = { workspace = true }
 unicode-segmentation = "1.10.0"
 peniko = { workspace = true }
-crossbeam-channel = "0.5.6"
+crossbeam-channel = { version = "0.5", optional = true }
 im-rc = "15.1.0"
 serde = { workspace = true, optional = true }
 lapce-xi-rope = { workspace = true, optional = true }
@@ -147,3 +147,5 @@ tokio = ["dep:tokio"]
 rfd-async-std = ["dep:rfd", "rfd/async-std"]
 rfd-tokio = ["dep:rfd", "rfd/tokio"]
 futures = ["dep:futures"]
+
+crossbeam = [ "dep:crossbeam-channel" ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,7 +58,6 @@ tokio = { version = "1", features = ["sync", "rt"], optional = true }
 raw-window-handle = { workspace = true }
 unicode-segmentation = "1.10.0"
 peniko = { workspace = true }
-crossbeam-channel = { version = "0.5", optional = true }
 im-rc = "15.1.0"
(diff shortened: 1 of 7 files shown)


src/ext_event.rs
```diff
@@ -158,6 +159,7 @@ pub fn update_signal_from_channel<T: Send + 'static>(
     });
 }
 
+#[cfg(feature = "crossbeam")]
 pub fn create_signal_from_channel<T: Send + 'static>(
     rx: crossbeam_channel::Receiver<T>,
```
panekj, line 164: Wouldn't it be better to put

```suggestion
    #[cfg(feature = "crossbeam")]
    rx: crossbeam_channel::Receiver<T>,
    #[cfg(feature = "crossbeam")]
    rx: std::sync::mpsc::Receiver<T>,
```

Cargo.toml
```diff
@@ -58,7 +58,7 @@ tokio = { version = "1", features = ["sync", "rt"], optional = true }
 raw-window-handle = { workspace = true }
 unicode-segmentation = "1.10.0"
 peniko = { workspace = true }
-crossbeam-channel = "0.5.6"
+crossbeam-channel = { version = "0.5", optional = true }
```
panekj, line 61: I think we can completely remove it since `crossbeam` re-exports whole `crossbeam-channel`

charlescgs: Noticed it as well, will change it.

src/ext_event.rs
charlescgs: Fixing now.

renderer/src/gpu_resources.rs
```diff
@@ -53,7 +56,11 @@ impl GpuResources {
         });
         // Channel passing to do async out-of-band within the winit event_loop since wasm can't
         // execute futures with a return value
+        #[cfg(feature = "crossbeam")]
         let (tx, rx) = channel::bounded(1);
+        #[cfg(not(feature = "crossbeam"))]
+        let (tx, rx) = sync_channel(1);
```
panekj, line 62: This could be more like, but I don't know if others agree to such style

```rust
#[cfg(feature = "crossbeam")]
use crossbeam::channel::{bounded as sync_channel, Receiver};
#[cfg(not(feature = "crossbeam"))]
use std::sync::mpsc::{sync_channel, Receiver};

[...]

        let (tx, rx) = sync_channel(1);
```

charlescgs: Was actually thinking about that solution :D

charlescgs:
@panekj I think it's good to go now, unless any other improvements can be done.

jrmoulton:
There has been some discussion on the discord about whether the std channels or crossbeam should be the default. I'm going to go ahead and merge this as is for now

PR #780, opened 2025-02-19
fix/improve animation reverse

jrmoulton:
Previously, animation in auto reverse would run at double time and when greater than 50% of the duration had passed, it would essentially run time backwards. 

This was fine with most easing functions but would lead to unnatural results when using Spring easing. 

I had previously built in a reverse_once that worked around this, when specifically opted into, but it makes sense for it to be how reversing is always done. Now time does not run backwards and animations don't run at double time. Both of these things are good changes. 

Now reversing all easing functions look natural.

### DIFF
--- a/src/animate.rs
+++ b/src/animate.rs
@@ -310,6 +310,7 @@ pub(crate) enum AnimState {
     // NOTE: If animation has `RepeatMode::LoopForever`, this state will never be reached.
     Completed {
         elapsed: Option<Duration>,
+        was_reversing: bool,
     },
 }
 
@@ -343,6 +344,8 @@ pub enum AnimStateCommand {
     Start,
     /// Stop the animation
     Stop,
+    /// Start the animation in reverse
+    Reverse,
 }
 
 type EffectStateVec = SmallVec<[RwSignal<SmallVec<[(ViewId, StackOffset<Animation>); 1]>>; 1]>;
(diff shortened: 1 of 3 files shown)


jrmoulton:
```rust
fn app_view() -> impl IntoView {
    let click = Trigger::new();
    let circle = |translate: bool| {
        empty()
            .style(move |s| {
                s.size(300., 300.)
                    .border(8.)
                    .border_radius(10.pct())
                    .background(Color::from_rgb8(144, 55, 48))
                    .border_color(Color::from_rgb8(195, 126, 144))
                    .border_progress(0.)
                    .translate_y(0.)
                    .rotate(0.)
                    .apply_if(!translate, |s| s.background(Color::from_rgb8(184, 187, 38)))
                    .absolute()
            })
            .animation(move |a| {
                a.keyframe(0, |k| k.computed_style().ease(Spring::bouncy()))
                    .keyframe(100, |k| {
                        k.ease(Spring::bouncy()).style(|s| {
                            s.apply_if(translate, |s| {
                                s.border_progress(100.).translate_y(150).rotate(2)
                            })
                        })
                    })
                    .duration(2000.millis())
                    .auto_reverse(true)
                    // .repeat(true)
                    .initial_state(animate::AnimStateCommand::Stop)
                    .start(move || click.track())
            })
            .on_click_stop(move |_| {
                click.notify();
            })
    };
    (circle(false), circle(true))
        .v_stack()
        .style(|s| s.size(300, 300))
        .container()
        .style(|s| {
            s.size_full()
                .items_center()
                .justify_center()
                .background(Color::from_rgb8(50, 48, 48))
        })
}
```

https://github.com/user-attachments/assets/f7531b5a-1db7-4689-be85-5f4b9dd0b983

PR #782, opened 2025-02-28
Support transparent windows in vger

jrmoulton:
This prefers either pre or post multiplied alpha modes if they are supported in the alpha capabilities. If neither is supported then it falls back to normal. 

In tiny skia this also supports transparent backgrounds but there are incorrect configurations in upstream softbuffer. if those were fixed then with this change it would just start working. 

vello is more complicated and I haven't been able to figure out what need to change to support transparent backgrounds. 

before merging this I want to make sure that this doesn't break things on windows or linux so I'd need someone to test those.


PolyMeilex:
Something went wrong with vger backend on Linux Wayland (tiny skia seems fine)

image

Interestingly enough, when I take a screenshot of a window rather than of the full composited frame, I can see transparent pixels on it (checkerboard added to make it more visible):

image
image

PR #790, opened 2025-03-11
Layer support

jrmoulton:
Add layer support to renderer (vello and tiny_skia)

This now also enables full view/layer effects such as opacity.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,6 +81,7 @@ im = { workspace = true }
 winit = { workspace = true }
 futures = { version = "0.3.30", optional = true }
 crossbeam = { version = "0.8", optional = true }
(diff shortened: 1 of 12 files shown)


PolyMeilex:
This seems to cause some regressions, eg. inspector looks like this for me:

image

Widget gallery is also missing some pieces of text:

image



Going back to `4bf974cc8a6ff43dd2e476595379ac91440ab89a` fixes the text rendering in the inspector (the inspector view is still blank tho, but that's probably caused by something else)

jrmoulton:
Hmm. This is probably related to window scale. What renderer is this?

PolyMeilex:
Either vger or vello, with Vulkan wgpu backend.

PR #791, opened 2025-03-14
fix tiny_skia text clip

jrmoulton:
@PolyMeilex can you test this?


PolyMeilex:
Yep, seems fine on all backends now.

PR #795, opened 2025-03-30
Fix: incorrect apply style order

charlescgs:
It fixed the problem of some styles not being applied: for example scroll handle thickness.


jrmoulton:
Thanks!

PR #803, opened 2025-04-22
Fix virtual_list Home keybind

PolyMeilex:
This was a fun one to debug :smile:

Because we use Any for update state, we have to specify the int literal type.


jrmoulton:
Oof thanks

PR #808, opened 2025-04-30
Fix text-input Chinese input

qingyiwebt:
Now IME is allowed for the window winit created, and the text input can handle the ImeCommit event,  
BUT the render issue when there is CJK text in the text-input still exists.

### DIFF
--- a/src/views/text_input.rs
+++ b/src/views/text_input.rs
@@ -138,11 +138,13 @@ pub fn text_input(buffer: RwSignal<String>) -> TextInput {
             id.update_state((text, is_focused));
             set_ime_allowed(is_focused);
 
-            let rect = id.layout_rect();
-            set_ime_cursor_area(
-                Point::new(rect.x0, rect.y0),
-                id.get_size().unwrap_or(Size::ZERO),
-            );
+            if is_focused {
+                let rect = id.layout_rect();
+                set_ime_cursor_area(
+                    Point::new(rect.x0, rect.y0),
+                    id.get_size().unwrap_or(Size::ZERO),
+                );
+            }
         });
     }
 


dzhou121:
We probably should call `set_ime_allowed(true)` when text input is in focus, and `set_ime_allowed(false)` when it's out of focus. 

https://github.com/lapce/floem/blob/main/src/action.rs#L213

qingyiwebt:
The calling of `set_ime_cursor_area` is also added in the commit [enable IME when text-input focused](https://github.com/lapce/floem/pull/808/commits/1681348c2ce6aad186ad00c6468ad53492bb8211)

qingyiwebt:
Now text-input should work fine with CJK characters and IME.

https://github.com/user-attachments/assets/ec1e8ce6-9d4d-4517-9c5b-bb8958eda4f1

image

src/views/text_input.rs
```diff
@@ -134,7 +134,15 @@ pub fn text_input(buffer: RwSignal<String>) -> TextInput {
     {
         create_effect(move |_| {
             let text = buffer.get();
-            id.update_state((text, is_focused.get()));
+            let is_focused = is_focused.get();
+            id.update_state((text, is_focused));
+            set_ime_allowed(is_focused);
+
+            let rect = id.layout_rect();
+            set_ime_cursor_area(
```
dzhou121, line 142: only do this when `is_focused = true`

src/views/text_input.rs
```diff
@@ -134,7 +134,17 @@ pub fn text_input(buffer: RwSignal<String>) -> TextInput {
     {
         create_effect(move |_| {
             let text = buffer.get();
-            id.update_state((text, is_focused.get()));
+            let is_focused = is_focused.get();
+            id.update_state((text, is_focused));
+            set_ime_allowed(is_focused);
```
dzhou121, line 139: There could be issue when the focus is switched from one text input to another, and we can't guarantee the execution order of the `set_ime_allowed`, so when the loosing focus text input execute this later than the gaining focus one, the ime would be disallowed. We need a central state of whether ime should be allowed.

PR #823, opened 2025-05-15
fix text selection

jrmoulton:
I agree that the default for Fleom should be non-selectable text by default as discussed in #822 but for now this just fixes the subtle issue of the events being eaten by requiring a minimum move distance

### DIFF
--- a/src/views/label.rs
+++ b/src/views/label.rs
@@ -333,10 +333,13 @@ impl View for Label {
                     else {
                         return EventPropagation::Continue;
                     };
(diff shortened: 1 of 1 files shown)


dzhou121:
I just tried on web that you can't select on buttons.

So I feel like we need to consume the poinerdown event if there's a on click event on the view.

jrmoulton:
I think this will have the desired effect. I don't think we want to handle the click event first.

dzhou121:
That makes sense. This minimum distance check is needed regardless. We can discuss/do #822 in a different PR/discussion.

PR #834, opened 2025-06-20
Implement cubic path segments, by approximating it to a quadratic curve

macmv:
This implements rendering cubic bezier curves, by approximating the curve to a quadratic.

There was some artifacting with the current `tolerance` of `0.1`, which went away when I bumped it up to `0.001`.

I tested this manually by drawing a circle segment:
```rs
let radius = 100.0;
let thickness = 3.0;

cx.fill(
  &CircleSegment {
    center:       Vec2::new(100.0, 100.0),
    outer_radius: (radius + thickness / 2.0).into(),
    inner_radius: (radius - thickness / 2.0).into(),
    start_angle:  0.0,
    sweep_angle:  std::f32::consts::FRAC_PI_3,
  },
  color,
  0.0,
);
```

I can post a standlone repro to test if you'd prefer, but right now I've just been testing in a larger app directly. Here's a picture of the result:

image

With the lower tolerance, I saw artifacting along the edges of the arcs:

image

### DIFF
--- a/vger/src/lib.rs
+++ b/vger/src/lib.rs
@@ -407,7 +407,7 @@ impl Renderer for VgerRenderer {
             )
         } else {
             let mut first = true;
-            for segment in path.path_segments(0.1) {
+            for segment in path.path_segments(0.001) {
                 match segment {
                     peniko::kurbo::PathSeg::Line(line) => {
                         if first {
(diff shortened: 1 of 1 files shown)


jrmoulton:
This looks good. thanks!

PR #841, opened 2025-06-23
Add pan-zoom view example

timsueberkrueb:
A simple pan-zoom area that supports translation and scaling.
My motivation of adding this example is twofold.
First, it can serve as a starting point for various types of visual editors that often need a pan-zoom area as a central part of their user interface.
Second, it is useful for testing/debugging rendering and event handling with transformations (scale, translation).


timsueberkrueb:
#842 is necessary for this example to work on Vger.

PR #843, opened 2025-06-23
vger: Text position should take window scale into account

timsueberkrueb:
Unfortunately, I introduced a bug in #842 in that the text position no longer took the window scale factor into account. This broke the `window-scale` example and rendering with Vger on a high dpi screens.


macmv:
looks good to me. thanks for the quick fix!

PR #845, opened 2025-06-24
Enhance pan-zoom example with clipping and images

timsueberkrueb:
Enhancing the pan-zoom example. This surfaces at least two existing bugs.

It can be observed that clipping to the rounded rectangle does not work, even though there is special logic for clipping rounded rectangles. The reason for this is that in `Context::clip`, we turn any rounded clip into a regular rectangle if an existing clip applies.

```rs
let rect = if let Some(existing) = self.clip {
    let rect = existing.rect().intersect(rect.rect());
    self.paint_state.renderer_mut().clip(&rect);
    rect.to_rounded_rect(0.0)
```
https://github.com/timsueberkrueb/floem/blob/28a4c724a9b6f843b0096e2d4aed168d415a1477/src/context.rs#L1161-L1162

Of course, intersecting rounded rectangles does not necessarily yield another rounded rectangle.

Further, zooming very close crashes Vger:

```
thread 'main' panicked at /home/tim/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wgpu-24.0.5/src/backend/wgpu_core.rs:1423:26:
wgpu error: Validation Error

Caused by:
  In Device::create_texture, label = 'atlas_texture'
    Dimension X value 14658 exceeds the limit of 8192


note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### DIFF
--- a/examples/pan-zoom/src/main.rs
+++ b/examples/pan-zoom/src/main.rs
@@ -1,5 +1,7 @@
 use floem::kurbo;
 use floem::prelude::*;
+use floem::style::StyleValue;
+use floem::style::TextColor;
 
 mod pan_zoom_view;
 mod transform_view;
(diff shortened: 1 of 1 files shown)


jrmoulton:
I've looked into the vger clipping bug and I don't think there is any way around. Vger just doesn't understand complex shapes so there isn't a clear way that those should be combined.

jrmoulton:
The atlas texture issue... we should probably cap the max scale/resolution that a glyph/svg can have when using vger

timsueberkrueb:
> The atlas texture issue... we should probably cap the max scale/resolution that a glyph/svg can have when using vger

I was wondering if clipping the generated images to the viewport before sending them to Vger might be sufficient? It seems we run into the limits because we're generating very large images but only chunks of them will actually fall inside the viewport. Ideally, we would also only render the parts of the images software-side that are visible. I don't know if this is feasible, though.

PR #863, opened 2025-08-03
fix: move popout menu to PointerUp event

panekj:
Fixes https://github.com/lapce/lapce/issues/3786

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -207,54 +207,63 @@ impl EventCx<'_> {
 
         // CLARIFY: should this be disabled when disable_default?
         if !disable_default {
-            match &event {
-                Event::PointerDown(event) => {
-                    self.app_state.clicking.insert(view_id);
-                    if event.button.is_primary() {
-                        let rect = view_id.get_size().unwrap_or_default().to_rect();
-                        let now_focused = rect.contains(event.pos);
+            let popout_menu = || {
+                let bottom_left = {
+                    let layout = view_state.borrow().layout_rect;
+                    Point::new(layout.x0, layout.y1)
+                };
 
-                        if now_focused {
+                let popout_menu = view_state.borrow().popout_menu.clone();
+                show_context_menu(popout_menu?(), Some(bottom_left));
+                Some((EventPropagation::Stop, PointerEventConsumed::Yes))
+            };
+
+            match &event {
+                Event::PointerDown(pointer_event) => {
+                    self.app_state.clicking.insert(view_id);
+                    if pointer_event.button.is_primary() {
+                        let rect = view_id.get_size().unwrap_or_default().to_rect();
+                        let on_view = rect.contains(pointer_event.pos);
+
+                        if on_view {
                             if self.app_state.keyboard_navigable.contains(&view_id) {
                                 // if the view can be focused, we update the focus
                                 self.app_state.update_focus(view_id, false);
                             }
-                            if event.count == 2
+                            if pointer_event.count == 2
                                 && view_state
                                     .borrow()
                                     .event_listeners
                                     .contains_key(&EventListener::DoubleClick)
                             {
-                                view_state.borrow_mut().last_pointer_down = Some(event.clone());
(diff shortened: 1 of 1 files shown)


jrmoulton:
at least on macOS the convention is for context menus to appear on pointer down events. I'm not sure this is the right fix

src/context.rs
```diff
@@ -207,54 +207,63 @@ impl EventCx<'_> {
...
+                                    Some(pointer_event.clone());
                             }
 
-                            let bottom_left = {
-                                let layout = view_state.borrow().layout_rect;
-                                Point::new(layout.x0, layout.y1)
+                            #[cfg(target_os = "macos")]
+                            if let Some((ep, pec)) = context_menu() {
+                                return (ep, pec);
                             };
```
jrmoulton, line 254: this should go under secondary click not primary, no?

jrmoulton: and same thing for not macOS

panekj: Look at the old diff

panekj: I guess the naming is wrong/confusing because it's not `context_menu` but `popout_menu` but the function used is actually `show_context_menu`?

panekj: https://github.com/lapce/floem/blob/67bb86ca17a387e8c00888fc7f79820894061172/src/context.rs#L242-L246

jrmoulton: ah yeah I was confused by the naming. so it should be on primary pointer down everywhere and probably called popout menu

jrmoulton: actually not sure about whether pointer down vs up or if there is a different convention on different platforms

panekj:
How does the [menu bar](https://developer.apple.com/design/human-interface-guidelines/the-menu-bar) behave?

jrmoulton:
> How does the [menu bar](https://developer.apple.com/design/human-interface-guidelines/the-menu-bar) behave?

It's pointer down

panekj:
https://learn.microsoft.com/en-us/windows/win32/uxguide/inter-mouse#simple-mouse-interactions

panekj:
| Simple action | Interaction | Typical effect |
| --- | --- | --- |
| Clicking | Position the pointer to a specific, non-selectable object and press and release a mouse button without moving. Clicking takes effect on the mouse button release to allow users the opportunity to cancel the click by moving the mouse off the target. Therefore, mouse press only indicates the selected target. | For single clicks with the primary button, activate the object. For double-clicks with the primary button, activate the object and perform the default command. For the secondary button, display the object's context menu. |

panekj:
Now I'm wondering if we should also implement clicking-cancel when taken pointer out of the object

panekj:
Also still trying to find similar guidelines in QT docs...

jrmoulton:
> Now I'm wondering if we should also implement clicking-cancel when taken pointer out of the object

I think it already is. if the mouse moves before the pointer up, then it won't receive the pointer up event to trigger the popout (unless the view specifically has requested to be active)

panekj:
Trying this in widget-gallery it behaves like that but I'm unsure if that is because the buttons are misbehaving (text is selectable)

panekj:
Context menu shows up even out of focus so it probably is correct

panekj:
Can I also rename `context_menu.rs` to `menu.rs` in widget-gallery to match it with sidebar in the gallery

panekj:
Changed `label()` to `button()` and now context menu doesn't actually show up when pointer leaves button

panekj:
But maybe this is correct according to Win32 guidelines? Since the context menu is attached to button target but if it was a standalone context menu without a target then it would just work

jrmoulton:
@panekj are you wanting this merged or do you want to wait?

panekj:
i think it's best to merge this for now (since it fixes the actual issue) while I'll go cry in a corner reading 15 different GUI toolkits documentations to find description of how mouse clicking works (because puters)

PR #868, opened 2025-08-28
Impl Default for RwSignal

macmv:
This comes in handy when I have a struct like:
```rust
#[derive(Clone, Copy, Default)]
struct Focuses {
  focus_left: RwSignal<()>,
  focus_right: RwSignal<()>,
  /* ... */
}
```

Maybe this isn't the best way to do this, but I couldn't figure out another way to handle `request_focus` easily.

### DIFF
--- a/reactive/src/signal.rs
+++ b/reactive/src/signal.rs
@@ -53,6 +53,12 @@ impl<T> fmt::Debug for RwSignal<T> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


jrmoulton:
Thanks! this looks like a good change but you might also consider using a Trigger (although I guess we should also add an implementation of default for trigger)

macmv:
oh, good idea. thanks!

Issue #867, opened 2025-08-28
virtual_stack: selected ViewID becomes obsolete as it leaves scroll view


bitfehler:
I hope the title makes sense, the issue can be observed e.g. in the `virtual_list` example (on current master):

- select an item
- scroll down until selected item out of view
- scroll back up

The selected item appears no longer selected. With my mighty println debugging, I determined that the virtual_stack's `selected` ViewId is something like `14v1`, but once scrolled out and back in, that ViewId never shows up anymore, but instead something like `14v5`, which of course never matches in the test whether to highlight this item.

I am afraid I understand too little about the ViewId internals to fix this, though I am happy to give it a stab if provided with a few pointers.

Or could this be an issue with the "view function", as implemented in virtual_list?

jrmoulton:
because the virtual list rebuilds elements as they go in and out of view, the view will lose focus when it is dropped. what we should probably do is have some way for the virtual list to track if any of it's views have focus, if they do track what it is by it's index, then when views are being rebuilt, if no other view in the app has requested focus and there is a view at an index that previously had focus, request focus on the newly built view id

jrmoulton:
maybe instead of just checking the same index it should check to see if the key hash matches

PR #869, opened 2025-08-29
virtual_stack: handle selection based on index

bitfehler:
The ViewId can be lost if the selected item is dropped from the underlying virtual vector, for example because of scrolling to a different part of the list. When the item gets recreated (when scrolling back to it), the ViewId will be different and the item will no longer be selected.

Conveniently, this allows dropping the `pending_idx` handling.

Fixes: #867

### DIFF
--- a/src/views/virtual_stack.rs
+++ b/src/views/virtual_stack.rs
@@ -82,8 +82,7 @@ where
     children: Vec<Option<(ViewId, Scope)>>,
     /// the index out of all of the items that is the first in the virtualized set. This is used to map an index to a [`ViewId`].
     first_child_idx: usize,
-    pending_idx: Option<usize>,
-    selected: HashSet<ViewId>,
+    selected_idx: HashSet<usize>,
     viewport: Rect,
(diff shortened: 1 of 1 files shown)


bitfehler:
I am not really sure if this is what you had in mind in your response to #867. For example, I do not understand what you meant by "it should check to see if the key hash matches". However, as is, it fixes the issue as I observed. However, I may well be unaware of certain behavior that I broke with this.

bitfehler:
Hm, I think I  get what you meant with the key now, comparing the result of the key function for that item to assert it is still the same. This seems a bit more involved,  so I'll think about how that could look like for a moment...

bitfehler:
The key function type is a bit of a pain to work with, tbh. The `+  Hash` makes it dyn-incompatible, so I cannot immediately come up with a way to make that available at the needed places that does not involve a giant amount of changes. If you have any ideas I'd love to hear them. I'll also keep thinking, though...

jrmoulton:
This looks good to me

PR #870, opened 2025-08-31
Fix clipping with window scale in vger

timsueberkrueb:
The clip rectangle needs to take the window scale factor into account. I changed to code s.t. the final text clipping calculations happen in the final coordinate space, i.e. with all transforms and the window scale factor (`VgerRenderer::scale`) being applied. Note that `VgerRenderer::scale` is not being used in other parts of the code besides `draw_text_with_layout`.

This bug was introduced by my earlier PR #842. I added comments that note the assumptions on the coordinate space currently in place.

I tested this PR manually on lapce (settings, viewing Rust code), the window scale example, the widget gallery and the pan-zoom example.


panekj:
It fixes the issue and is extremely small so I'm merging this

PR #875, opened 2025-09-15
feat: add ranged slider

alettsy:
Hello 👋

First of all, thank you for an awesome Rust package! 😊

## Summary
I have added an additional slider type (**ranged**) so that it is easier to define a slider between certain bounds.

## Ranged slider
The slider can be created using the constructor `slider::Slider::new_ranged(...)`.
It expects a closure which returns the `f64` value (not percentage) and an inclusive range.

For example,
```Rust
slider::Slider::new_ranged(move || value.get(), -50.0..=150.0)
```

## `range` and `step`

I have added two new properties: `range` and `step`.

The `range` defines the minimum and maximum values the slider should be clamped between.
The `step` defines by how much each value change should increase/decrease the value by.

When dragging the slider, it will automatically snap values to the nearest `step`.

By default, the `range` is set to `0.0..=100.0` and the `step` is set to `1.0`.

Note that `step` only effects the actual value (output by `on_change_value`), not the percentage and pixel values.

## `on_change_value(...)`

I have added a new function to emit the actual value the slider is sitting on, rather than the pixel shift or percentage.
It can be set up in the same way as the other two "on change" functions.

The value emitted will be within the defined `range` and it will follow the chosen `step`.

## `step(f64)`

The step can be modified by a function on the slider.

For example,
```Rust
slider::Slider::new_ranged(move || value.get(), -50.0..=150.0).step(10.0)
```

I kept it this way so that the step can also be applied to the regular sliders (without a range).

## Examples

I have provided an example application with all the different ways to create a slider under `examples/slider-types`. Hopefully it should provide a clear idea of what I was trying to achieve with this new addition!

## Final note

I have not really contributed to Rust projects before, so please let me know if there are any issues or bad practise things that you notice so I can improve the code!

Thank you very much! 😊


jrmoulton:
Thanks for the PR! This looks like a good change but I think it needs to rework some of the internal details for consistency. I think it would be good for the slider to always use the range when calculating the value and should just default to 0..100. and I think it can always use the step as well, the step should just be optional. and I think we should only keep the on change px and on change pct and not add the on change value

alettsy:
Thank you very much for the feedback! I appreciate the quick response 😊

I can certainly make the `step` optional. 

I think the goal of adding the `on_change_value` method is to provide a way for the user to get the actual value that the slider represents, rather than just the percentage or pixel position of the handle, eliminating the need for manual conversions to usable values.

If we kept only the other two methods, then the user can only get the percentage and pixel position of the handle from the slider, so if the value they wanted is different than `0.0..=100.0`, then they would have to calculate it themselves based on the percentage emitted. The new `on_change_value` method does it for them instead, using the `range` to map the percentage to a value.

For example, it would eliminate the need to map the values yourself, like this:
```Rust
let percent_value = create_rw_signal(0.pct());

// [slider code omitted]

let range = -50.0..=150.0;

let range_diff = range.end() - range.start();
let actual_value = range.start() + (range_diff * (percent_value.get() / 100,0));

// with on_change_value, the value emitted is already mapped and stepped for you to use straight away
// so you can just use my_value.get()
```

From what you suggested, we could instead have the slider default to `0.0..=100.0` and internally calculate the value based on the provided `range` and `step`, then convert the value to the percentage for the `on_change_pct` method. But this is mostly the same as what we are doing now, just from value to percent, rather than percent to value, and it doesn't allow the user to access the value, only the percentage and the handle position, unless we keep the `on_change_value` method too.

Please let me know what you think! 😊

alettsy:
It looks like the checks for clippy failed because of something unrelated to my PR? 

If you want, I can change the problematic lines from:
`assert!(bytes_per_row % wgpu::COPY_BYTES_PER_ROW_ALIGNMENT == 0);`
to
`assert!(bytes_per_row.is_multiple_of(wgpu::COPY_BYTES_PER_ROW_ALIGNMENT));`

to fix this error 😊

jrmoulton:
yeah this is because of the new rust release from a few days ago, it's failing with a new clippy lint. 

> If you want, I can change the problematic lines from:

Yes please

jrmoulton (approved):
This looks good

alettsy:
> This looks good

Thank you! 😊

I have added the fix for the clippy errors too

jrmoulton:
Thanks!

PR #880, opened 2025-09-26
integrate ui-events

jrmoulton:
This would need switched to using the upstream ui events crate. 

I had an experiment with making the ui events crate generic over the pointer type but it's probably not a good idea. Switching to the ui events crate means that we will have dpi types as part of events which are part of the public api. 

We should have a very easy way to get a kurbo Point from a dpi type. An extension trait might be what we have to do


jrmoulton:
I've made a smaller change to upstream ui events that I think will be accepted. This pr currently depends on a branch of mine of ui-events. 

I've copied the ui-events-winit crate into floem and updated it to work with our fork of winit. 

This is a big tedious change that will break users of floem but using ui-events will enable us to have more control of our events and allow us to have adapters to windowing backends like base view.

PR #882, opened 2025-09-30
Improve support for Futures, Streams, Channels with custom executors

jrmoulton:


jrmoulton:
@josie-mshrm 

I've changed this to implement a from any type that can be turned into a futures stream. Does this cover more of your use cases?

josie-mshrm:
oh that's super cool! i'm not the one doing async though, i'm using crossbeam with threads and i think that was covered before?

jrmoulton:
> oh that's super cool! i'm not the one doing async though, i'm using crossbeam with threads and i think that was covered before?

Ah sorry. Pinged the wrong person. 

@yiffyrusdev

jrmoulton:
I've done an overhaul of Resource and added a FutureSignal, StreamSignal, and a ChannelSignal and made it so that you can select the executor that drives the future/stream/receiver
One of the provided executors is the event loop, so no Send bound and tokio not required (as long as the future doesn't need the tokio reactor)

yiffyrusdev:
Sorry, i was waaay to far. We had a chat in ds anyway
I just hopped here to say - this is SUPER cool!

PR #884, opened 2025-10-02
Make overlays direct children of the window's root view

### LINKED ISSUE #883: Dropdown selection issues
Dropdowns with more items than the window can fit make it impossible to select some items with the mouse.

There are three issues at play causing this problem:
- [OverlayView offsets render](https://github.com/lapce/floem/blob/6442b3f22192e1cca77e531e6406db1998a3bd9e/src/window_handle.rs#L1654) when the overlay can't fit on screen, which causes the true layout to not align with the render.
- The dropdown is a child of an element that can't restrict its size, preventing scrollbars from appearing.
- The list inside of the `scroll()` has a `size_full` style, which also seems to prevent scrollbars from appearing.
  - I switched this to `width_full` while playing with possible fixes, and it allows scrollbars to appear if the previous issue is also fixed.

Screenshot of the relevant issues:
Image

Expecting it to look something like this:
Image

I made the sample image with this modification to OverlayView, along with some other changes:

```rs
struct OverlayView {
    id: ViewId,
    child: ViewId,
    position: Point,
}

impl View for OverlayView {
    fn id(&self) -> ViewId {
        self.id
    }

    fn view_style(&self) -> Option<crate::style::Style> {
        Some(
            Style::new()
                .absolute()
                .padding_left(self.position.x)
                .padding_top(self.position.y)
                .padding_bottom(5)
                .size_full()
                .pointer_events_none(),
        )
    }

    fn debug_name(&self) -> std::borrow::Cow<'static, str> {
        "Overlay".into()
    }

    fn paint(&mut self, cx: &mut PaintCx) {
        cx.paint_view(self.child);
    }
}
```

I think simplifying overlays further or adding a new method to directly place views as children of the window would make this easier to fix and more flexible for new types of overlays that may want to play with alignment instead of rendering at a specific position.

ArthurCose:
Fixes #883 

Changes `add_overlay` to directly accept a view and return that view's id, rather than creating an intermediate overlay view between the window root view and the view passed to `add_overlay`. This allows styles on overlays to work in the context of the root view rather than an unsized intermediate.

Dropdowns avoid rendering off-screen using styles:
image

Tooltips still stay on screen with a specialized styled intermediate:
image

### DIFF
--- a/src/action.rs
+++ b/src/action.rs
@@ -23,6 +23,7 @@ use crate::{
     menu::Menu,
     update::{UpdateMessage, UPDATE_MESSAGES},
     view::View,
+    views::Decorators,
     window_handle::{get_current_view, set_current_view},
 };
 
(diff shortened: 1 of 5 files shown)


jrmoulton:
Thanks!

Issue #403, opened 2024-03-28
Editor selection no longer shows selection on empty lines


MinusGix:
![image](https://github.com/lapce/floem/assets/13157904/5203c00e-98db-4c02-945e-6d4ae91a7014)

Probably due to no longer including newlines in the visual line, so selection logic needs some +1 to size if the lines are empty.

KhodyrevDS:
fixed https://github.com/lapce/floem/commit/b2fe244aca8128332ffcef38c882e931d0acdd1b

PR #892, opened 2025-10-03
Use normal flow to render empty selections

ArthurCose:
Not certain on this change as I'm unsure on what this branch was intended for.

Before change, newline selection renders at the start of the line rather than the end:
image

After:
image

### DIFF
--- a/src/views/editor/view.rs
+++ b/src/views/editor/view.rs
@@ -373,14 +373,6 @@ impl EditorView {
 
             let line_height = f64::from(ed.line_height(line));
 
-            // Skip over empty selections
-            if !info.is_empty_phantom() && left_col == right_col {
-                let rect = Rect::from_origin_size((0.0, vline_y), (CHAR_WIDTH, line_height));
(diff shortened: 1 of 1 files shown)


jrmoulton:
#397 and #403/#708 are in the blame for history

ArthurCose:
I believe these are preserved after this change:

https://github.com/user-attachments/assets/64e4791d-a39d-4ea4-9f28-531773cff41e

ArthurCose:
Line 5 has a space, the rest of the blank appearing lines really are blank for clarity.

ArthurCose:
I'm going to dig a bit further, looks like there's more history.

ArthurCose:
I'm guessing this is relevant:
image

I'm only selecting the number `1` on the top right on this wrapped line. Left is before my change, right is after. It's a little broken in both versions, but I'll give fixing this a shot.

ArthurCose:
This is accounted for now:

https://github.com/user-attachments/assets/1e5a560f-6589-494c-802b-f63b26b8c1e0

PR #893, opened 2025-10-03
Fix text editor not growing to new size without update

ArthurCose:
Issue fixed by this change:
image

Reproducing the issue:
Resize only the width of the window in the editor example. The text editor will only grow if an edit is made in the editor or the window is resized vertically.

### DIFF
--- a/src/views/editor/view.rs
+++ b/src/views/editor/view.rs
@@ -1201,6 +1201,7 @@ fn editor_content(
                     }
                 }
             })
+            .style(|s| s.min_size_full())
     })
     .on_move(move |point| {
         window_origin.set(point);


jrmoulton:
Looking at the original port of of the editor from Lapce to Floem [here](https://github.com/ArthurCose/floem/blame/f0e12db2ddd86d58cce408411c4a7960e423ba7f/src/views/editor/view.rs#L1169) this was included. 

Looks like I accidentally deleted it [here](https://github.com/ArthurCose/floem/blame/fbd5f4db12db93b3fcbe047df6f6cc054e8ff634/src/views/editor/view.rs#L1228).

PR #894, opened 2025-10-03
Fix cursor rendering for CursorAffinity::Backward

ArthurCose:
Before change, cursor affinity appears ignored outside of selections. When there is a selection, the cursor renders on the wrong line:

https://github.com/user-attachments/assets/11e55e69-1924-4fea-9395-25b6cb2019e2

After change:

https://github.com/user-attachments/assets/c2328405-2ad2-46ad-937f-1e390f61c1ac

### DIFF
--- a/src/views/editor/movement.rs
+++ b/src/views/editor/movement.rs
@@ -274,14 +274,7 @@ fn move_right(
         }
     }
 
-    let (rvline, col) = view.rvline_col_of_offset(offset, *affinity);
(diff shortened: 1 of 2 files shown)


jrmoulton (approved):
I've dug into this a bit and this looks correct to me and it now matches the move left behavior where move left always sets the affinity to forward and move right always now sets the affinity to Backward regardless of if it's at the end of the line. 

@dzhou121 does this look correct to you?

PR #898, opened 2025-10-04
don't request paint on every style change

jrmoulton:
This fixes a 2 year old todo where a paint is was being requested if any style changed. 

Views should be responsible for requesting paint when needed. 

This allows for making updates to styles that don't affect the visual appearance without needing to always repaint. 

@panekj 
If there is any stuttering in custom views in Lapce, those may need to make a call to `ViewId.request_paint()` when they have updates

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -637,6 +637,8 @@ impl<'a> StyleCx<'a> {
         CaptureState::capture_style(view_id, self, computed_style.clone());
         view_state.borrow_mut().computed_style = computed_style;
 
+        let old_view_style_props = view_state.borrow().view_style_props.clone();
+
         // This is used by the `request_transition` and `style` methods below.
         self.current_view = view_id;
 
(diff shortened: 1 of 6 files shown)


dzhou121:
What are the cases where a style change doesn’t need repaint?

i feel like this approach will put a  cognitive burden. 

Maybe we could associate the repaint logic with specific styles?

jrmoulton:
There are lots of cases. The style system can be used to pass arbitrary properties to a view. Disabling text selection only changes how the view responds to events and the only place with the context to know if that should trigger a repaint or not is the label view. 

I've added an automatic repaint request any time the view style properties change so most views are mostly unaffected. But there are a few cases where views need to request repaint on state change. 

I feel like it is in line with the rest of the view trait for the view to be in charge of making those types of decisions because it is the only place with the right context.

jrmoulton:
marking as a draft for now

umbrageodotus:
How would requesting paints manually be ergonomic? Maybe each style could define repaint criteria?

PR #899, opened 2025-10-04
Give a CursorAffinity to each selection region

ArthurCose:
Fixes issues with multi-cursors at line wrapping boundaries shifting from using a shared affinity.

Library API change summary:

- `SelRegion` now has an affinity property.
- `SelRegion` constructors and `Selection` shorthand constructors (such as `Selection::region`) now require CursorAffinity.
- `CursorMode` and `Cursor` now have an `affinity()` method for grabbing the affinity of the latest region.
- `CursorMode::Normal` and `CursorMode::Visual` now have an affinity property.
- `Cursor::set_offset` and `Cursor::add_region` now require CursorAffinity.
- `Cursor::regions_iter` now includes CursorAffinity as the third value in its tuple.

Before on left, after on right:

https://github.com/user-attachments/assets/6ede17b4-370d-4b0f-8404-b2466fdf7a30

I'm not 100% sure on the correctness of every change involved, as I'm not deeply familiar with relevant features of the text editor, such as the CursorModes and every command. I think at worst, for the issues I'm worried about (aside from whether the changes at their core seem good) will just mean in some cases a cursor may appear on the wrong line from what users expect, which can be caught and fixed over time.

I'd definitely like to test out the different CursorModes if I could have guidance on that, or if someone could take a look for me.

### DIFF
--- a/editor-core/src/buffer/test.rs
+++ b/editor-core/src/buffer/test.rs
@@ -4,14 +4,20 @@ mod editing {
     use lapce_xi_rope::Rope;
 
     use super::*;
-    use crate::{editor::EditType, selection::Selection};
+    use crate::{cursor::CursorAffinity, editor::EditType, selection::Selection};
 
     #[test]
     fn is_pristine() {
         let mut buffer = Buffer::new("");
         buffer.init_content(Rope::from("abc"));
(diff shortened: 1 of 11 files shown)


jrmoulton (approved):
If there are no objections, I'll merge this tomorrow.

PR #903, opened 2025-10-07
Update to Vello 0.6

charlescgs:
Updated to newest vello, run cargo: clippy, update & fmt.

@jrmoulton should we do anything with the new `depth_slice` field on `RenderPassColorAttachment`? I filled this witth `None` and it seems to work just fine.


jrmoulton:
using `..Default::default()` is probably better but `None` is fine.

PR #913, opened 2025-10-10
Prevent multi cursor overlap

ArthurCose:
Left is before change, right is after:

https://github.com/user-attachments/assets/9f0d84d3-65cc-414e-b3e9-61cd76c515d3

Window on the left crashed with `Delta builder: intervals not properly sorted` after attempting to type something


ArthurCose:
Switched to draft after finding a new crash

PR #917, opened 2025-10-11
Render placeholder text to regular text layout, simplify clip logic

### LINKED ISSUE #196: Placeholder overflows text input
~When the input is set as a percentage and can be changed with the window size it may overflow the input itself as shown in the screenshot below.~

Edit: When the placeholder is longer than the `text_input` it will always overflow regardless of what width the `text_input` is set to

image

ArthurCose:
Fixes #196 

I'm aiming to simplify `text_buf` as just the visual buffer and not map it 1:1 with the real buffer. Aside from allowing the placeholder text to clip, this should open it up to embed the IME preedit text.

https://github.com/user-attachments/assets/82cefc55-880b-4c4d-9505-139a9ec46f19


jrmoulton:
I'll probably get around to reviewing this tonight

PR #920, opened 2025-10-12
Update theme - Please test / try out / give opinions

jrmoulton:
WIP theme update. 

image
image
image
image
image
image
image

### DIFF
--- a/examples/themes/src/main.rs
+++ b/examples/themes/src/main.rs
@@ -4,8 +4,8 @@ use floem::{
     peniko::{color::palette, Color},
     reactive::{create_signal, SignalGet, SignalUpdate},
     style::{
-        Background, BorderBottomColor, BorderLeftColor, BorderRightColor, BorderTopColor, Outline,
-        OutlineColor, Style, TextColor, Transition,
+        Background, BorderColorProp, BorderProp, Outline, OutlineColor, Style, TextColor,
+        Transition,
     },
     style_class,
     unit::DurationUnitExt,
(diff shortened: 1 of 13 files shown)


timsueberkrueb:
Looking great!

Here are some things I noticed:

image

Highlighting in dropdown (`widget-gallery`) doesn't seem to change with selection.

image

Disabled but selected radio button "Windows" (`widget-gallery`) looks odd. Looks fine in dark mode, though.

image

There are odd white corners in dark mode in the "DroppedFile" example (`widget-gallery`).

jrmoulton:
> Here are some things I noticed:

Thanks! Should be fixed now

jrmoulton:
This is close to being ready to merge but I need to find some way to do robust resolution of nested style maps. Currently support for some nested maps is supported but it's never clear what will or won't work.

jrmoulton:
This does now do recursive style map resolution. Somewhere along the way though I have made styling 10x slower... which is way too much. It's happening in `Style::apply_mut`. Style maps must be getting combined way more often than before and I need to find a way to reduce the number of times maps get combined and hopefully reuse maps

jrmoulton:
Ok it's actually only about 2x slower. which is still too much but not as bad as I thought. Need to find some way to track if a nested map actually has useful properties

jrmoulton:
This ended up being a massive PR. Styling with this complicated theme is naturally slower than before but I have made it so that full restyles are much less likely to happen and so, in general, styling performance should be improved overall. If you don't use complicated themes, styling should be just as fast as before. I have tried to carefully check that everything is working but it is likely that some things have broken. I also haven't implemented every suggestion to the default theme. For theme updates please open PRs to fix those things and for any bugs please open a PR or an issue.

PR #921, opened 2025-10-12
Improve vello performance and reliability

jrmoulton:
@execsuroot can you try this out?


jrmoulton:
This is a bit of an ugly hack but it works significantly better for vello. When we switch to using the anyrender abstraction, it will support vello style clips better and the code will look more consistent between renderers.

PR #944, opened 2025-10-26
Fix: Text input selection paint size

charlescgs:
This PR is adjusting text input selection paint size to cover better selected text and avoid clipping left edge of the selection:

### Examples
|     | Ex. 1 | Ex. 2 | Ex. 3 | Ex. 4 |
|--|--|--|--|--|
| **Before** |  | |  |  |
| **After** |  |  |  |  |

### DIFF
--- a/examples/widget-gallery/src/main.rs
+++ b/examples/widget-gallery/src/main.rs
@@ -77,9 +77,7 @@ fn app_view(window_id: WindowId) -> impl IntoView {
             _ => label(|| "Not implemented".to_owned()).into_any(),
         }
         .debug_name(it.to_string())
-    }.style(|s| s
-        .scale(150.pct())
-    );
+    };
 
     let tabs = RwSignal::new(tabs);
 


jrmoulton:
Does the text layout provide any other information that we should be using to better position the selection? just inflating the rect feels like a solution that will cause as many problems as it solves

charlescgs:
Hmm, this also seems to work:
```rust
let selection_rect = self
    .get_selection_rect(&node_layout, padding_left + 1.)
    .to_rounded_rect(border_radius);
cx.save();
// cx.clip(&self.id
//     .get_size()
//     .unwrap_or(Size::new(selection_rect.width(), selection_rect.height()))
//     .to_rect()
// );
cx.fill(&selection_rect, &cursor_color, 0.0);
cx.restore();
```
Clip does appear to do nothing here though..  🤔

jrmoulton:
It makes sense to have the clip in case there are rounded corners or some case where it would clip. 

I think we do want to actually fix this properly using info from the TextLayout

charlescgs:
Copy that, but I don't feel competent enough to do that, have looked briefly there and have no idea what is going on :D

One note on the side; this change affects only `text_input` widget, but not the editor, so scope for anything going wrong is very limited, I have tested this change quite extensively on every case I could think of and did not see any regress.

jrmoulton:
I think it used to have an inflate on it and I removed it some time ago because it was causing problems

PR #948, opened 2025-10-27
Add localization

charlescgs:


jrmoulton:
@charlescgs

jrmoulton:
sorry for the noise. ended up removing im since im_rc is preferred

jrmoulton:
also for safety I think I need to pin the struct since I am hacking lifetimes to allow for self referential stuff

jrmoulton:
@charlescgs I moved the module to be called localization

PR #956, opened 2025-11-01
Apply transform when deciding whether to propagate events

timsueberkrueb:
`EventCx::should_send` determines if an event should be propagated to a (child) view. Previously, this function didn't take the style transforms into account. This means, for example, that clicking the button in the `pan-zoom` example didn't have an effect if the transformed content had been moved away from its starting position.

With this PR, events should now be propagated properly even if style transformations have been applied.

<details>

[Screencast From 2025-11-01 17-12-29.webm](https://github.com/user-attachments/assets/2233feee-78f3-454c-88bd-f4f777572d87)

</details>

As far as I can tell, it seems to match the logic in `EventCx::offset_event` now; please double-check!

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -515,6 +515,9 @@ impl EventCx<'_> {
             layout.location.x as f64,
             layout.location.y as f64,
         ));
+        // For that, we need to take any style transformations into account
+        let transform = id.state().borrow().transform;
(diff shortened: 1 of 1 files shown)


jrmoulton:
This looks good. I'll do some manual testing and then merge this later today

PR #957, opened 2025-11-01
Remove trailing whitespace and use LF over CRLF

timsueberkrueb:
This PR removes trailing whitespace from all text files. It also harmonizes the line-endings in `src/nav.rs` and `examples/webgpu/fonts/FiraSans-LICENSE` from `CRLF` to `LF` for consistency with the rest of the project.


jrmoulton:
Trailing whitespace in markdown will produce a line break. The trailing whitespace in markdown should either be left alone or replaced with an explicit line break.

timsueberkrueb:
Thanks, I wasn't aware. I added newlines where relevant. Btw, I'm also happy for this to be closed if you think it's not worth the trouble or if you prefer whitespace for newlines in Markdown. The trailing whitespace just annoyed me as it sometimes causes churn in my diffs with my editor configured to trim it.

jrmoulton:
This looks good. Thanks!

PR #959, opened 2025-11-05
Feat: make theme change apply to all windows

charlescgs:
This PR adds:
- theme change to apply to all windows
- way to override theme for specific window
- changes from #960 with fixed window menu behavior
- new window to respect overridden theme (if set)
- three theme related actions for better ergonomics

|Before | After |
|--------|-------|
|  https://github.com/user-attachments/assets/ce21c8a6-1ac7-4f95-97fd-9e312b0a2a56 | https://github.com/user-attachments/assets/0cd15212-b4d5-4c65-8c06-33398fb32e6c |

### DIFF
--- a/examples/widget-gallery/src/main.rs
+++ b/examples/widget-gallery/src/main.rs
@@ -225,6 +225,7 @@ fn app_view(window_id: WindowId) -> impl IntoView {
             .item("Set Dark Theme", |i| {
                 i.action(|| set_theme(Some(Theme::Dark)))
             })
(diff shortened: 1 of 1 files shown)


jrmoulton:
I don't think we want the default behavior of set theme to set the theme for all windows. I think that should be a separate update. I think a per window theme/override should be possible

charlescgs:
Copy, will separate those features, will use code from #960.

jrmoulton:
added a follow os theme option

jrmoulton:
Thanks!

PR #960, opened 2025-11-05
fix follow os theme

jrmoulton:
@charlescgs this makes it so that the window will properly follow changes to the theme from the OS if not overridden.

### DIFF
--- a/src/action.rs
+++ b/src/action.rs
@@ -25,7 +25,6 @@ use crate::{
     view::View,
     views::Decorators,
     window_handle::{get_current_view, set_current_view},
-    window_tracking::with_window,
 };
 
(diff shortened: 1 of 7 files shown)


jrmoulton:
@charlescgs could you test this on windows? before clicking the toggle theme button, all windows should follow OS theme updates multiple times (previously it was only working once). Then once the toggle button is pressed, it should act as an override and then theme updates from the OS should be ignored

jrmoulton:
@charlescgs are you still planning on integrating this into #959?

charlescgs:
Yes, sorry for delay, but got really busy week at work, will update PR tomorrow.

jrmoulton:
Ok sounds good. No worries

jrmoulton:
closed in favor of #959

PR #961, opened 2025-11-05
async rfd

jrmoulton:
Now that it is easier to run futures on the event loop I have moved the file dialogs to use the new `FutureSignal`. Now file dialogs no longer need to spawn a thread


jrmoulton:
This needs testing on linux and windows. Just checking that file dialogs in the widget gallery files example will be sufficient

- [ ] linux
- [x] windows
- [x] macOS

ArthurCose:
Works for me on Linux, but I do get a lag spike / pause before the dialog appears, which differs from what I see on the main branch.

jrmoulton:
That's strange. I'll probably just mark this as draft and leave it then. According to the rfd docs, this is better for macOS, but as it works just fine. Thanks for testing

charlescgs:
on Windows it work just as well as before.

PR #963, opened 2025-11-05
fix tab nav

jrmoulton:
@charlescgs


charlescgs:
Can confirm with change fixes the issue!

PR #964, opened 2025-11-06
Refactor async signal

jrmoulton:
@execsuroot hopefully improved async/channel/stream api that allows for an initial value so that the getters are not always `Option<T>`


jrmoulton:
there is probably some way to reduce code duplication with this but after a few attempts I just accepted it for what it is. I like the API

execsuroot:
> @execsuroot hopefully improved async/channel/stream api that allows for an initial value so that the getters are not always `Option<T>`

Will fix breaking changes in my code today and test it out!

PR #965, opened 2025-11-07
fix style schedule

jrmoulton:
When resolving  nested maps, if a class is applied, we need to request style recursive on the children


jrmoulton:
~~hmmm... this broke the widget gallery pretty bad...~~

fixed. I had replaced a request to layout with a request to style accidentally.

PR #969, opened 2025-11-10
rely on request redraw for event processing

jrmoulton:
this is necessary. if we try to process events right after they happen, if the processing takes a long time, there will be a long string of queued up events before the request redraw and we will miss frames. 

We should as much as possible ensure that events emitted by winit other than request redraw do not block.


jrmoulton:
This will be something that anyone reviewing prs should be mindful of and as we make changes to Floem, try to reduce blocking everywhere that isn't under a request redraw.

jrmoulton:
This is currently breaking timers so things like exec_after aren't running. the timers are being created but they are never being fired

jrmoulton:
I'm reworking the layout system. I'll get back to this once that is done in the next day or so

PR #972, opened 2025-11-11
`DesignSystem` improvements

charlescgs:
This PR is making an effort to improve upon existing `DesignSystem`.
- Set up lightness and muted parameters on per theme basis instead relying on hard-coded values
- Add some documentation
- Add `hsl` helper function

### DIFF
--- a/src/theme.rs
+++ b/src/theme.rs
@@ -34,12 +34,9 @@ pub struct DesignSystem {
     pub bg_elevate: f32,
     /// Lightness of the background overlays from 0-1.
     pub bg_overlay: f32,
-    /// Lightness of the border (0-1) based on background.
-    pub border: f32,
-    /// Theme border radius.
-    pub border_radius: f32,
     /// Lightness of the disabled elements (0-1) based on background.
     pub disabled: f32,
+
     /// Base text color.
     pub text_base: Color,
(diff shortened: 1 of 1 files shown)


charlescgs:
Bumped to 1.90 as `floor` was constified at this rust version.

jrmoulton:
Hmm. I thought that this would improve the performance of the style pass but it doesn't actually seem to make a difference. 

What was the motivation for this change?

There is a regression on disabled text color

jrmoulton:
image

jrmoulton:
> While we use Color that is just an alias for AlphaColor<Srgb> that is all I can do.

Styles in most places accept a `Brush` which can be any color space (or a gradient/image)

charlescgs:
The motivation was to make it do more work at compile time instead of runtime, but if its not making a difference then yeah, feel free to close it.

jrmoulton:
I think all of the clippy if let chains would be good to merge. Maybe let's just revert the changes to theme.rs

charlescgs:
Will do that. However, will keep `hsl` function as an alternative way of constructing color, which might be useful when you are trying your theme out. Will add also a way to customize background elevation level.

charlescgs:
@jrmoulton PR is ready for review.

PR #980, opened 2025-11-17
Extract `view_style` from #973 & make `default_compute_layout` public

charlescgs:
I believe `.default_compute_layout()` should be public as when you try to implement `View` trait for a custom view, you still sometimes want to call that method (as it is the case for our dropdown implementation in example).

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -642,18 +642,30 @@ impl<'a> StyleCx<'a> {
         let view_state = view_id.state();
         {
             let mut view_state = view_state.borrow_mut();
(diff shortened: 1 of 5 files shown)


jrmoulton:
In the layout overhaul I'm doing the default_compute_layout will be gone and it will be significantly easier to do advanced layout. I have built in more direct ways of interacting with taffy, using custom measure functions with taffy, getting layout in local coordinate space, parent coordinate space, window coordinate space and made all of that very fast. I am also building in ways to interact with the final visual layout tree so you can implement highly efficient hit testing in your own views. but this is a fine change for now

PR #985, opened 2025-11-27
add SyncRwSignal for cross thread usage

dzhou121:
Added SyncRwSignal for cross thread usage, and the current RwSignal is still a local version. 

SyncRwSignal will enqueue the Effect runs to UI thread. 

Prefer struct constructors now and deprecate the free functions.


jrmoulton (approved):
Didn't review every detail but this looks great. This will be nice to have

PR #992, opened 2025-12-04
add `with_win_os_config` method to `WindowConfig`

alvinisspicy:
What does this PR do?

- Adds a `with_win_os_config` method to `WindowConfig` to expose `WinOSWindowConfig` through the public builder API.

- Mirrors the existing `with_mac_os_config` and `with_web_config` pattern so Windows-specific window options can be configured ergonomically.

Why is this needed?

- `WindowConfig` already stores an optional `WinOSWindowConfig` that is applied during window creation on Windows, but there was no public way for users to modify it.

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -240,6 +240,23 @@ impl WindowConfig {
         self
     }
 
+    /// Set up Windows specific configuration. The passed closure will only be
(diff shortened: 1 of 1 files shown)


jrmoulton:
Thanks!

PR #993, opened 2025-12-05
fix broken DurationUnitExt::minutes for u64

alvinisspicy:
Fixes the incorrect implementation of DurationUnitExt::minutes for u64.


jrmoulton:
Thanks for fixing this!

PR #997, opened 2025-12-18
Add z-index aware event dispatch

dzhou121:
Implement CSS-like stacking context behavior for pointer events:
- Events now dispatch to views based on visual z-index order, not DOM order
- Higher z-index views receive events before lower z-index siblings
- Stacking contexts are properly bounded (child z-index doesn't escape parent)

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -859,8 +859,7 @@ impl<'a> StyleCx<'a> {
         let position = view_state.borrow().combined_style.get(PositionProp);
         let has_transform = transform != Affine::IDENTITY;
 
-        let creates_context =
-            z_index.is_some() || position == Position::Absolute || has_transform;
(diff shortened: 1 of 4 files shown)


jrmoulton:
I have some changes that will mostly overwrite this. I have redone the event handling and it is now significantly more robust and respects z index and also has explicit phases for capture, target and bubble. It also does hit testing using a spatial index so no more need to slowly walk the view tree for hit testing. In the changes I have, events are still dispatched in dom order, but the event context caries the visual hit path so an event handler can see/use the visual order if desired

dzhou121:
Do you want to push it and let's take a look?

jrmoulton:
draft in #998. The event handling is in src/context/event.rs. needs a bit more work to make propagation work

PR #998, opened 2025-12-18
Integrate understory

jrmoulton:
This has a bunch of changes that all came as a result of integrating understory but that's made it quite a large change

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -104,10 +104,13 @@ wgpu.workspace = true
 fluent-bundle = { version = "0.16", optional = true }
 unic-langid = { version = "0.9", optional = true }
 sys-locale = { version = "0.3.2", optional = true }
+understory_index = {path = "../understory/understory_index/"}
 understory_box_tree = {path = "../understory/understory_box_tree/"}
 understory_responder = {path = "../understory/understory_responder/", features = ["box_tree_adapter"]}
 understory_focus = {path = "../understory/understory_focus/", features = ["box_tree_adapter"]}
 understory_precise_hit = {path = "../understory/understory_precise_hit"}
+understory_event_state = {path = "../understory/understory_event_state"}
+thiserror = "1.0"
 
 [target.'cfg(any(target_os = "windows", target_os = "macos"))'.dependencies]
 muda = { workspace = true }
@@ -177,3 +180,4 @@ localization = ["dep:fluent-bundle", "dep:unic-langid", "dep:sys-locale"]
 
 [patch.crates-io]
 ui-events = {path = "../ui-events/ui-events/"}
+
--- a/examples/context/src/main.rs
+++ b/examples/context/src/main.rs
@@ -4,7 +4,7 @@ use floem::{
     reactive::{provide_context, use_context},
     ui_events::keyboard::{Key, NamedKey},
     views::{empty, label, v_stack, Decorators},
-    IntoView, View,
+    IntoView,
 };
 
 fn colored_label(text: String) -> impl IntoView {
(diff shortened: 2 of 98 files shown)


jrmoulton:
the biggest change here is event handling. there is a new src/context/event.rs file that does the new event handling. It does the dispatch of pointer events using understory_box_tree which allows for fast hit testing. The box tree is updated after doing taffy layout and it integrates any visual transformation (scale/rotation/scroll offset) so that we get correct hit testing for targets that have transformed after layout. 

because scroll offsets are now treated as a visual transform along with the others I needed to rework both the scroll and virtual list widgets. 

While working on event handling I have made it so that event handlers get both mutable access to an EventCx and also mutable access to the View. This is largely why so many files have changed. 

I have also reworked layout to be push based. So there is no layout pass, views need to push changes to the taffy tree (which is not mostly done automatically by Floem) and request layout or push properties into the box tree and request box tree commit. Views can request a function run post layout if needed but this is not run by default. 

to do removal of the layout pass I needed to rework text layout so that we use the custom taffy measure functions. This has a nice benefit of making it so that layout with line breaking can run in a single pass and we don't have to round trip multiple times to the view update and back to layout. This makes a noticeable improvement to paint latency, so now when resizing the window, even in debug mode, when there is lots of linebreaking, we still repaint before the frame deadline.

Styling is now cached so dirty tracking is done in the window state and an explicit style traversal is created to only style the necessary views. 

paint has changed as well a bit (there are some bugs I just haven't fixed yet) so that they get the final visual position correctly from the box tree

dzhou121:
I can't see understory in crates.io. How are we going to publish floem?

jrmoulton:
> I can't see understory in crates.io. How are we going to publish floem?

@waywardmonkeys are we good to do some initial releases soon? 

I don't think there will be any pushback to that

https://github.com/endoli/understory

jrmoulton:
important for me, conceptually this enables the separation of a "visual ID" from a View. We can use box tree node ids as a visual id for a rectangle on the screen and support hit testing, interaction states, event targeting, and updates for those rectangles with identifiers without needing to have the full overhead of a View. 

https://xi.zulipchat.com/#narrow/channel/550935-understory/topic/Floem.20Integration/near/563959235

I wrote a bit here on a nice pattern this enables. 

this is useful for any app that wants to control the layout or lifecycle of elements on the screen without needing to reimplement a whole ui library just to get hit testing and interaction states and final screen positioning

jrmoulton:
Currently I am using main branch understory and [this](https://github.com/endoli/understory/pull/61) open pr. I'd actually like to move away from needing that PR though. currently I need it because we only have one box tree per app, in the same way we only have one taffy tree per app. but for performance reasons we should really have one spatial index/box tree per window. I haven't done this because when view state is built, it isn't yet determined which window/root that view belongs to so we can't know which box tree should be providing the node id. We could initialize it lazily but that felt a bit ugly, it might be best though. 

other things that aren't done / need fixed

- [ ] Not all event listeners are inserted 
- [ ] event propagation is not being stopped
- [ ] event cx should have a way of preventing default actions
- [ ] on pointer events we should be storing an unfiltered hit path in the eventcx. We should try to limit the number of times we hit test and instead just filter on the single unfiltered path multiple times. 
- [ ] the paint pass needs fixed. ideally we will use the box tree to drive an explicit traversal of visual ids that respect z index. 
- [ ] text overflow listeners are not hooked up
- [ ] resize event handlers are not hooked up. There should be a nice way to use the damage rects from the box tree commit and a set of elements that have those listeners to efficiently update those listeners. 
- [ ] we should have an explicitly requested `post_box_tree_commit` pass similar to `post_layout`. 
- [ ] overlays should respond to changes in visual positioning so they stay pinned to their views even during scroll or the view transforming.
- [ ] working through a few bugs where sometimes the style traversal isn't completely cleared. Might be fixed now needs more testing
- [ ] styles are not being requested properly on window resize for responsive styles
- [ ] we should probably switch the spatial index that the box tree is using to the grid backend
- [ ] adopt understory selection
- [ ] virtual list isn't updating on first paint. while fixing that we might as well use understory virtual list as it should work well for us
- [ ] view drag isn't being handled. it should be much easier to do correctly now that we can get world_bounds and world_transform for local->window coordinate space conversion.

PR #1004, opened 2025-12-23
add as_any_mut to View trait

jrmoulton:
This is used as part of some changes to event handlers. I have made it so that event handlers get mutable access to the view. That will come in a follow up PR.

### DIFF
--- a/examples/pan-zoom/src/pan_zoom_view.rs
+++ b/examples/pan-zoom/src/pan_zoom_view.rs
@@ -172,6 +172,10 @@ impl View for PanZoomView {
             self.schedule_update();
         }
     }
+
+    fn as_any_mut(&mut self) -> &mut dyn std::any::Any {
+        self
+    }
 }
 
 impl PanZoomView {
--- a/examples/pan-zoom/src/transform_view.rs
+++ b/examples/pan-zoom/src/transform_view.rs
@@ -113,6 +113,10 @@ impl View for TransformView {
         cx.compute_view_layout(self.child_id);
         None
     }
+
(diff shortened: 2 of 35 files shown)


dzhou121:
Is it a good idea to let user have mutable access to view? Floem generally update stuff via message/event

jrmoulton:
I think it's a good idea. It's just the exact same as having mutable access to the view in the `event` methods of the View trait. public/private can still be controlled by widgets but this let's you do anything you could do in the normal event methods from event handlers

dzhou121:
This is for the event listeners to have mutable access to view I believe?

jrmoulton:
> This is for the event listeners to have mutable access to view I believe?

yes

dzhou121:
So for a user to update the view, we'd prefer it sends a message rather than directly mutation.

jrmoulton:
I get that for a lot of views the update message will be the chosen api but that can be controlled with private fields. not allowing mutable access is just an arbitrary limitation. 

web allows for mutation in event handlers using `this`. 

Why force the usage of an update message?

dzhou121:
My main concern is preserving unidirectional data flow. Floem's architecture is: state changes → messages → view updates. This makes it easy to trace how the UI reached any given state. as_any_mut introduces a second path where views can be mutated directly, bypassing this flow. React, Vue, Elm, and SwiftUI all converged on unidirectional patterns because bidirectional mutation didn't scale. It becomes hard to reason about state as apps grow. If there's a specific use case that's painful with messages, let's discuss a targeted solution.

PR #1005, opened 2025-12-24
cache styles and use an explicit style traversal

jrmoulton:
This caches view styles and tracks view style dirtiness in hash maps in the window state. We then use those dirty maps to create an explicit traversal for styling (sorting the dirty nodes based on dfs). A style context is created for that specific view and the context is built using the relevant cached state and then the view style is computed and the style pass is run. 

A potential optimization is to track the `self.style()` stack using a `dirty_from` index and cache so that if there are many calls to .style on a view, only the minimal number of them is recomputed.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -30,9 +30,10 @@ use crate::dropped_file::FileDragEvent;
 use crate::easing::{Easing, Linear};
 use crate::menu::Menu;
 use crate::renderer::Renderer;
+use crate::responsive::ScreenSizeBp;
 use crate::style::{
-    Disabled, DisplayProp, Focusable, Hidden, PointerEvents, PointerEventsProp, PositionProp,
-    ZIndex,
+    Focusable, PointerEvents, PointerEventsProp, PositionProp, StyleClassRef, ZIndex,
+    resolve_nested_maps,
 };
 use crate::view_state::{IsHiddenState, StackingInfo};
 use crate::{
@@ -879,6 +880,8 @@ pub struct InteractionState {
     pub is_selected: bool,
     /// Whether this view is disabled.
     pub is_disabled: bool,
+    /// Whether this view is hidden.
+    pub is_hidden: bool,
     /// Whether this view has keyboard focus.
     pub is_focused: bool,
     /// Whether this view is being clicked (pointer down but not yet up).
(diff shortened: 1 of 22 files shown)


jrmoulton:
~~I also removed the transform handling from the style pass. it wasn't correct. It needs to be done using values from the final layout. The way I have it implemented in the understory PR is correct.~~ Added back for now

src/view_state.rs
```diff
@@ -196,13 +195,17 @@ pub struct ViewState {
...
     pub(crate) animations: Stack<Animation>,
     pub(crate) classes: Vec<StyleClassRef>,
     pub(crate) dragging_style: Option<Style>,
     /// Combine the stacked style into one style, and apply the interact state.
     pub(crate) combined_style: Style,
     /// The final style including inherited style from parent.
     pub(crate) computed_style: Style,
+    pub(crate) style_cx: Option<Style>,
+    pub(crate) style_interaction_cx: InheritedInteractionCx,
+    pub(crate) parent_set_style_interaction: InheritedInteractionCx,
```
jrmoulton: Need to add docs.

test/tests/scroll_behavior.rs
```diff
@@ -570,6 +570,7 @@ fn test_scroll_after_resize() {
     // Scroll down
     harness.scroll_down(50.0, 50.0, 150.0);
     let _scroll_before = scroll_tracker.last_viewport().unwrap().y0;
+    dbg!(_scroll_before);
```
jrmoulton: remove

src/context.rs
```diff
@@ -891,92 +894,139 @@ pub struct InteractionState {
...
+    pub(crate) disabled: bool,
+    pub(crate) hidden: bool,
+    pub(crate) selected: bool,
+}
+// impl InheritedInteractionCx {
+//     fn clear(&mut self) {
+//         self.hidden = false;
+//         self.disabled = false;
+//         self.selected = false;
+//     }
```
jrmoulton: delete

src/id.rs
```diff
@@ -326,27 +341,32 @@ impl ViewId {
...
     /// Returns true if the computed style for this view is marked as hidden by setting in this view, or any parent, `Hidden` to true. For hiding views, you should prefer to set `Hidden` to true rather than using `Display::None` as checking for `Hidden` is cheaper, more correct, and used for optimizations in Floem
     pub fn is_hidden(&self) -> bool {
+        self.state().borrow_mut().style_interaction_cx.hidden
+    }
+
+    /// if the view has pointer events none
+    pub fn pointer_events_none(&self) -> bool {
         let state = self.state();
         let state = state.borrow();
-        state.computed_style.get(Hidden) || state.computed_style.get(DisplayProp) == Display::None
```
jrmoulton: add back the hidden check

jrmoulton:
this also disables 
```rust
[profile.dev]
opt-level = 1
```
in the cargo toml that I enabled a bit after the theme update because of poor performance. it is now good enough that this can be removed but there are still significant performance gains to be had in the understory integration.

PR #1022, opened 2026-01-23
input method: Implement deletion in text input

dcz-self:
Deletion was missing completely. Now it works.

Tested with [stiwri](https://codeberg.org/dcz/stiwri/src/branch/sync2) and the corresponding fork of anvil, because that's what I can quickly prepare now. But it should work without the need for any experimental protocols. Phosh would be a good teting environment.

[Screencast_20260123_192312.webm](https://github.com/user-attachments/assets/f3651420-c089-4ecf-95fc-b4c5b6c37a8e)

### DIFF
--- a/src/event/mod.rs
+++ b/src/event/mod.rs
@@ -92,6 +92,8 @@ pub enum EventListener {
     ImeDisabled,
     /// Receives [`Event::ImePreedit`]
     ImePreedit,
+    /// Receives [`Event::ImeDeleteSurrounding`]
+    ImeDeleteSurrounding,
     /// Receives [`Event::ImeCommit`]
     ImeCommit,
     /// Receives [`Event::PointerWheel`]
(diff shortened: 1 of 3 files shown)


dcz-self:
I tried to add support to the text area, but I seem to be getting outdated cursor position: https://github.com/dcz-self/floem/tree/ime2

jrmoulton:
Looking at this code I don't think it will support grapheme clusters properly. I'm not sure how smart the delete ime event is but I think in the case that the delete is in the middle of a grapheme cluster we need to find the start and delete the entire cluster. The unicode crate is available

dcz-self:
That's correct. Generally, what gets deleted depends on the input method.

The case where the index falls inside a code point is actually (mistakenly) foridden by the protocol and will only happen when multiple sources are typing in the same spot faster than the input method can respond.

So I'd say basically never.

I'm not sure handling that rare case warrants the additional complexity. Should I implement grapheme clusters anyway?

jrmoulton:
>Should I implement grapheme clusters anyway? 

No this sounds reasonable. 

Could you add that as a comment in the code?

jrmoulton:
you can ignore the ci errors. they are from a new rust version. I'll merge and fix it after the comment is added

dcz-self:
Done and done.

jrmoulton:
Thanks!

PR #1026, opened 2026-02-06
fix wasm32

jrmoulton:
I haven't actually tested the that the context menus on wasm32 work but this builds so this is at least better


jrmoulton:
I do want to find a better solution for consistent imports that there aren't so many cfgs around but I don't think I'll do that in this PR

PR #1029, opened 2026-02-06
Fix IME text deletion

dcz-self:
The text was deleted, but the cursor stayed at its old index. Oops.


dcz-self:
Another fix: the current selection should be deleted when present. That's not mandated by winit because of an oversight: I forgot to include the explanation. But deleting selection is mandated in the Wayland protocol at least.

PR #1035, opened 2026-03-06
Inspector box model debug view

jrmoulton:
image

### DIFF
--- a/src/style/recalc.rs
+++ b/src/style/recalc.rs
@@ -348,8 +348,10 @@ impl StyleReason {
         let mut out = self.clone();
         out.targets.clear();
         out.flags.remove(StyleReasonFlags::TARGET);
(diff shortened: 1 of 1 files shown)


jrmoulton:
Oh and actually more important in this PR is that I am removing the style prop combining. I had done it originally because I wanted to combine the debug views for those elements into a group but combining them makes it so that cannot be transitioned independently. so I am removing it and I added a debug group that can be added to the style that will group apply a custom grouping debug view. I feel like this is moderately important because we don't have any other way to ensure related properties are shown together and having them all split makes it hard to understand which ones are active or not.

jrmoulton:
The box model view was largely done by codex. It mostly just worked. the style of code isn't my favorite but for a debug view I was mostly just happy that AI could produce working floem code without too much of a struggle

PR #1041, opened 2026-03-10
Faster style

jrmoulton:
Not nearly as big of a difference as I was hoping. It's about 12% faster

### DIFF
--- a/src/style/values.rs
+++ b/src/style/values.rs
@@ -1383,7 +1383,7 @@ fn short_style_name(name: &str) -> String {
 }
 
 struct StyleDebugRow {
-    render: Box<dyn FnOnce(bool) -> AnyView>,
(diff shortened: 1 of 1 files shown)


jrmoulton:
Just using FxHashMap is about 20-25% faster...

jrmoulton:
This is basically the same as the im hash map but just an `Rc<FxHashMap>`. it's about 20% faster. I also like that it's not the im hash map. 

when scrolling in the virtual list styling is still a bit slow but not a major bottle neck. I now get about 50fps in debug mode while scrolling through a million items. (up from 30). Bringing back the style cache would still be good. 

~~I think after that the only big win left will be resolving styles in parallel. I think it should be pretty easy to do. I think if we just wrap view state in Mutex but still guarantee that the style_pass runs on the main thread that will be the last big win in styling.~~ actually parallelizing style doesn't look so easy

PR #1044, opened 2026-03-13
add skia

jrmoulton:
the implementation is a bit ugly but it's also not that bad. and probably won't live for long and can be improved if it does. 

@execsuroot


jrmoulton:
I'm waiting on some prs on anyrender but I'm going to just go ahead and merge for now. I expect this entire implementation to be reworked but I want to land and then iterate on it.  I think I'll also add a vello cpu backend, try and get some benchmarks in place and then try to switch to a retained/incremental rendering abstraction.

PR #1045, opened 2026-03-14
Fix style context

jrmoulton:
This removes the context expressions that built entire maps and instead replaces them with expressions that operate per property. This fixes correctness issues that the map version had. the issue before was that when merging style maps, we merge per property. but context expressions with their nested properties would skip over any merges and get applied at the end meaning you couldn't override properties that had been set in a context expression. The solution is to have an expression that is stored per property so that context expressions get follow the normal rules for being overridden. this is a bit more unwieldy to use than before but it's reasonable with helpers. for the most part with helpers the code doesn't change at all. 

the thing this does not do is the especially late resolution of context like the theme getting access to a font size that is set in a local style. The only useful case I had for this before was font size but the answer to this is to just commit to a unit overhaul and add em units (which has the other benefit of correctly using transitioned values). I have that in another change that I'll push soon.

### DIFF
--- a/examples/todo-complex/src/todo.rs
+++ b/examples/todo-complex/src/todo.rs
@@ -205,7 +205,7 @@ impl IntoView for TodoState {
                     AppCommand::SetSelected(self).execute();
                 }
             })
-            .style(|s| s.width_full().align_items(Some(AlignItems::FlexStart)));
+            .style(|s| s.width_full().align_items(AlignItems::FlexStart));
 
         let container = main_controls.container();
         let final_view_id = container.id();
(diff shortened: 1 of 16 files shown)


jrmoulton:
The biggest thing that I don't like is needing to mirror style methods (mostly in macro but a few need to be done explicitly) for easy Into<> values. without that its very ugly to use. I tried a few different things but none worked well. so I ended up with the ExprMap which is a wrapper similar to the BuiltinMap that wraps style and provides different methods. that does work fine but it doesn't solve custom style methods... I wish there was a better way to get custom style methods to be done automatically. maybe there is a way to have the prop extractor macro do that.

PR #1048, opened 2026-03-16
adds try_doc to give people the Option

Third-Thing:
See https://github.com/lapce/lapce/pull/3876 for context.

Instead of the framework forcing a crash on dropped UI elements, this addition gives the host application the agency to handle them gracefully via an Option. This allows Lapce to safely iterate over editor components without needing defensive `catch_unwind` wrappers to protect against any desync between Lapce's editors map and Floem's reactive state.

### DIFF
--- a/src/views/editor/mod.rs
+++ b/src/views/editor/mod.rs
@@ -323,7 +323,15 @@ impl Editor {
         self.doc.get_untracked()
     }
 
+    /// Retrieves the [Document] and subscribes the current reactive scope to updates.
+    /// Returns `None` if the underlying reactive scope has been disposed.
     pub fn try_doc(&self) -> Option<Rc<dyn Document>> {
+        self.doc.try_get()
+    }
+
+    /// Retrieves the [Document] without subscribing to reactive updates.
+    /// Returns `None` if the underlying reactive scope has been disposed.
+    pub fn try_doc_untracked(&self) -> Option<Rc<dyn Document>> {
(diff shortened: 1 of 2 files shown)


jrmoulton:
This doesn't look like a bad change but it looks like it's working around an underlying problem. What is the situation that this is happening in?

Third-Thing:
The steps to produce the crash in lapce are:

1. open/close settings
2. create new document
3. file → save all

The underlying problem was solved in that PR by https://github.com/lapce/lapce/pull/3876/changes/7a61a550a14d491965f0c93248380b3cdd7e3e6e
but the concern was that plugins could hypothetically forget to add such clean up for their own content. Then, anyone using such a plugin would experience a similar crash caused by the Save All loop, hence the `catch_unwind` safety net (https://github.com/lapce/lapce/pull/3876/changes/5feb090a5d2a9a52e4e85d470201d511987d70f7).

`try_doc` would allow for app developers to handle the problem idiomatically:

```rust
let Some(doc) = editor_data.try_doc() else {
                tracing::warn!(
                    "Ghost editor bypassed during Save All. Editor ID: {:?}", 
                    editor_id
                );
                continue;
            };
```
rather than using `catch_unwind` to suppress floem's internal panic.

Third-Thing:
I've made some changes. Let me know if I interpreted your request correctly.

jrmoulton (approved):
Looks great! Thanks!

Issue #1051, opened 2026-03-20
Seemingly no way to intercept window close from application code


Third-Thing:
The doc comments on `Event::Window` describe `CloseRequested` and `Destroyed` as separate events, but as far as I can tell, the `WindowEvent` enum only has `Closed`. In `handle_window_event()`, winit's `CloseRequested` goes straight to `close_window()` → `handle.destroy()` with no event dispatched to the view tree beforehand.

This means application code can't intercept the close for things like unsaved-changes prompts or cleanup logic.

winit already separates `CloseRequested` from `Destroyed` for exactly this reason. It looks like Floem intended the same separation (based on the doc comments) but it wasn't completed.

Would it make sense to add a `CloseRequested` variant to `WindowEvent`, dispatch it before calling `close_window()`, and skip the close if the event is consumed? If so, I can start the work after your approval.

---

Status Update: `Completed`

jrmoulton:
yeah that makes sense to add and the semantics seem good.

PR #1053, opened 2026-03-22
adds WindowCloseRequested event and request_close_window()

Third-Thing:
Addresses issue #1051 

Adds `WindowCloseRequested` event with `prevent_default()` support,
`request_close_window()` as the interceptable counterpart to `close_window()`, 
and an example demonstrating a confirmation dialog.

I have thoroughly reviewed and tested these changes.


Third-Thing:
Good idea. The bool return did seem like a bit of a hack. I'll work on the changes.

jrmoulton:
Thanks!

PR #1055, opened 2026-03-27
fix offset_to_line test failure

Third-Thing:
The test hardcoded expected vline numbers (`4`, `5`) that assumed character `d` at offset 16 would land exactly at a soft wrap boundary at width 36px. ~Hypothesis: These values depended on specific font metrics from the pre-Parley text layout engine. After the Parley integration, SansSerif at 12pt renders with slightly different character widths, shifting the wrap point so `d` no longer starts a new visual line.~

~Rather than adjusting the magic numbers to match Parley's metrics (which would break again on the next layout engine or font change),~ the fix discovers the actual wrap boundaries at runtime and asserts the property the test was always meant to verify: at a soft wrap boundary, `Forward` affinity resolves to the next visual line and `Backward` resolves to the previous one. A guard assertion ensures the test fixture still produces at least one soft wrap, so it can't pass vacuously.

### DIFF
--- a/src/views/editor/visual_line.rs
+++ b/src/views/editor/visual_line.rs
@@ -2919,18 +2919,39 @@ mod tests {
                 .get(),
             1
         );
-        // starts at 'd'. Tests that cursor affinity works for soft line breaks
+        // Find a real soft-wrap boundary and verify that affinity selects
+        // the previous/next visual line at that shared buffer offset.
+        let wrapped_line = 4;
+        let wrapped_rvlines: Vec<_> = lines
+            .iter_rvlines(&text_prov, false, RVLine::new(wrapped_line, 0))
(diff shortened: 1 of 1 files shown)


Third-Thing:
On second glance it looks like the Parley PR added tests, so probably didn't miss a failing one.

Another potential cause could be https://github.com/lapce/floem/commit/ec9024f40a51df869613f4e2d713a0603f310199 changing two things that directly affect wrapping in the test:

1. **Wrap mode**: `Wrap::Word` → `TextWrapMode::Wrap` + `OverflowWrap::BreakWord` 
2. **Call order**: the old code set wrap mode *before* `set_text()`; the new code calls `set_text()` *first*, then applies wrap settings after

PR #1056, opened 2026-03-29
Add smooth scroll to the scroll view.

t800-x:
Added a `smooth_scroll` style that animates the scroll ticks with a linear animation. Had to move the `set_positions()` call out of `apply_scroll_delta()` cause while animating the tree is being modified and read at the same time, and it was causing understory_box_tree to panic and crash. 

I tried to use exponential decay instead of a linear curve but had problems chaining scroll ticks in quick succession. Did the best I could.

### DIFF
--- a/src/views/scroll.rs
+++ b/src/views/scroll.rs
@@ -2,7 +2,7 @@
 //! Scroll View
 
 use floem_reactive::Effect;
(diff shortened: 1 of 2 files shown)


jrmoulton:
Exciting! Can you instead let the style prop be a `DirectTranstion` instead of just a bool. that should help clean up the state management as well

t800-x:
Alright mate, done that. Also fixed a clippy warning in `tiny_skia/src/lib.rs`

PR #1062, opened 2026-04-04
enable undo cursor restoration for programmatic edits

Third-Thing:
Problem: Programmatic document edits currently bypass the editor context that  undo cursor restoration depends on. `Document::edit` and `edit_single`  mutate the buffer and emit deltas, but they do not record  `cursor_before` / `cursor_after`, so undo lacks the initiating editor state needed to restore the cursor correctly. That left floem’s examples  using raw document edits from button callbacks and compensating with workarounds like disabling undo in one shared editor.

Solution: Add additive editor-aware document edit APIs: `edit_from` and  `edit_single_from`. `TextDocument` uses the initiating editor to capture cursor-before state, apply the edit, remap that editor’s cursor through the produced deltas, and record cursor-after state on the revision. Raw  `edit` / `edit_single` remain unchanged as low-level document mutation  APIs, while UI-driven edits can now preserve expected cursor and undo behavior without changing the broader document-centered reactive model.

Reasoning: This keeps the architecture aligned with floem’s existing update flow.  Live cursor consistency across editors still comes from document deltas and update listeners; the new APIs only add what undo needs. The examples are moved to the editor-aware path, which removes the need to suppress undo in the secondary editor and makes the intended API split explicit.

Associated with issue https://github.com/lapce/floem/issues/1058

### DIFF
--- a/examples/editor/src/main.rs
+++ b/examples/editor/src/main.rs
@@ -1,10 +1,7 @@
 use floem::{
     prelude::*,
     views::editor::{
-        command::{Command, CommandExecuted},
-        core::{
-            command::EditCommand, cursor::CursorAffinity, editor::EditType, selection::Selection,
-        },
+        core::{cursor::CursorAffinity, editor::EditType, selection::Selection},
         text::{default_dark_color, SimpleStyling},
     },
 };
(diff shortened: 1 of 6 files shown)


Third-Thing:
This PR is stacked on top of #1061, because it relies on the test block added in that branch. That is why the first two commits are included. This is a consequence of using Gitbutler for the first time, not intentional.

PR #1074, opened 2026-05-11
recover from lost/outdated wgpu surface on Wayland

inanc-g:
On some Wayland compositors like Niri, floem apps freeze on startup. 

The first frame renders, then the window is visually frozen.  
Inputs still work under the hood but nothing visually changes.  
Resizing the window fixes it.  

To reproduce, simply run any example on niri (I haven't tried any other WM). 
-> Window is frozen, on resize, it works again.

This is same bug Zed had: [Freezing on Start in Niri #50574
](https://github.com/zed-industries/zed/issues/50574)
Their fix on their internal `gpui_wgpu` renderer: [Fix handling of surface.configure on Linux#50640
](https://github.com/zed-industries/zed/pull/50640)

In floem, for the vello and vger renderers, `finish()` doesn't handle any Err Result from `surface.get_current_texture()`. On some Wayland compositors, the surface might be "lost" or "outdated" right after initial configuration, so `get_current_texture()` returns an error on the very first frame. 

`SurfaceError` in wgpu doesn't auto-recover, so all future frames also fail silently and the window appears frozen.

Resizing unfreezes because the resize calls `surface.configure()` and rebuilds.

The fix is simply to handle `SurfaceError::Lost and SurfaceError::Outdated` by calling surface.configure() again, basically what Zed did:

```rust
Err(wgpu::SurfaceError::Lost | wgpu::SurfaceError::Outdated) => 
 self.surface.configure(&self.device, &self.config)
```

Note that this is also the cause of an open bug in Lapce: [Lapce freezing on niri at startup #3897
](https://github.com/lapce/lapce/issues/3897)

### DIFF
--- a/vger/src/lib.rs
+++ b/vger/src/lib.rs
@@ -710,29 +710,35 @@ impl Renderer for VgerRenderer {
         if self.capture {
             self.render_image()
         } else {
-            if let Ok(frame) = self.surface.get_current_texture() {
-                let texture_view = frame
-                    .texture
-                    .create_view(&wgpu::TextureViewDescriptor::default());
-                let desc = wgpu::RenderPassDescriptor {
-                    label: None,
-                    color_attachments: &[Some(wgpu::RenderPassColorAttachment {
-                        view: &texture_view,
-                        resolve_target: None,
-                        ops: wgpu::Operations {
-                            load: wgpu::LoadOp::Clear(wgpu::Color::TRANSPARENT),
-                            store: StoreOp::Store,
-                        },
-                        depth_slice: None,
-                    })],
-                    depth_stencil_attachment: None,
-                    timestamp_writes: None,
-                    occlusion_query_set: None,
-                };
+            let frame = match self.surface.get_current_texture() {
+                Ok(frame) => frame,
+                Err(wgpu::SurfaceError::Lost | wgpu::SurfaceError::Outdated) => {
+                    self.surface.configure(&self.device, &self.config);
+                    return None;
+                }
+                _ => return None,
+            };
+            let texture_view = frame
+                .texture
+                .create_view(&wgpu::TextureViewDescriptor::default());
+            let desc = wgpu::RenderPassDescriptor {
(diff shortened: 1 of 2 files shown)


vger/src/lib.rs
```diff
@@ -710,29 +710,35 @@ impl Renderer for VgerRenderer {
...
-                    timestamp_writes: None,
-                    occlusion_query_set: None,
-                };
-
-                self.vger.encode(&desc);
-                frame.present();
-            }
+            let frame = match self.surface.get_current_texture() {
+                Ok(frame) => frame,
+                Err(wgpu::SurfaceError::Lost | wgpu::SurfaceError::Outdated) => {
```
0xNULLderef, line 715: [WebGpu examples](https://github.com/gfx-rs/wgpu/blob/af91efa93da2af9366b252ed7b9a089175f4874e/examples/features/src/framework.rs#L229) also choose to handle `wgpu::SurfaceError::Other` and `wgpu::SurfaceError::OutOfMemory`. Not relevant on niri (from my testing), but it might help getting a present instead of a frozen mess in some other specific edge cases

they also choose to attempt a get again but I'm not that thoroughly versed in Vulkan to decide whether this or the other approach is better

inanc-g: That's interesting. In the `wgpu` docs, it's described a bit differently:

`Lost` and `Outdated` are the two variants where `wgpu` itself explicitly recommends recreating the surface: https://docs.rs/wgpu/27.0.1/wgpu/enum.SurfaceError.html#variants

It seems in their example they opted for a broad "just reconfigure will fix it" catch-all approach, but it might make sense to only handle what's actually a surface configuration issue.

OOM shouldn't really try to recreate the surface if memory is starved (wouldn't be able to create a new frame anway - the comment in their example even says that), and `Other` would need to be inspected to decide what to do.

jrmoulton (approved):

jrmoulton:
This looks good. Thanks!

PR #1075, opened 2026-05-11
Rework rendering

jrmoulton:
This isn't quite done. I still need to add windows and linux and web support for the native compositor stuff in subduction (which probably also involves a lot of refactoring and simplification). 

There is also a new frame pacing crate that I have in understory but I might move it to subduction. 

Things are generally fast in this code but there is a lot of new code an opportunity for improvement. 

I don't plan on waiting for review on understory/imaging/subduction as the review time is pretty unworkable. I'll just get the changes into my fork and get things merged here.


jrmoulton:
I'll do a bigger write up of the changes

PR #1079, opened 2026-06-04
Fix wasm compilation/rendering

Fabi019:
Thank you for this great UI library! While trying it out in a WASM environment I found and fixed some small issues:

- Compilation errors due to some non wasm compatible types used (`muda::Menu`, `std::time::Instant`)
- No render output due to `self.paint_state` not getting set to `Initialized` without paint called
- Invisible text on label/buttons in the `webgpu` example -> Manually set the `font_family` to "Fira Sans"

### DIFF
--- a/src/window/handle.rs
+++ b/src/window/handle.rs
@@ -786,15 +786,12 @@ impl WindowHandle {
     }
 
     pub(crate) fn render_frame(&mut self) {
-        #[cfg(not(target_arch = "wasm32"))]
         let renderer_ready = matches!(self.paint_state, PaintState::Initialized { .. });
-        #[cfg(target_arch = "wasm32")]
-        let renderer_ready = true;
         if self.window_state.request_paint && renderer_ready {
             self.window_state.request_paint = false;
             self.paint();
-            self.last_presented_at = Instant::now();
         }
+        self.last_presented_at = Instant::now();
 
         if self.live_resize_active() {
             self.window_state.schedule_paint(self.id);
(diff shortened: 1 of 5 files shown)


umbrageodotus:
LGTM

src/window/handle.rs
```diff
@@ -786,7 +786,10 @@ impl WindowHandle {
     }
 
     pub(crate) fn render_frame(&mut self) {
+        #[cfg(not(target_arch = "wasm32"))]
         let renderer_ready = matches!(self.paint_state, PaintState::Initialized { .. });
+        #[cfg(target_arch = "wasm32")]
+        let renderer_ready = true;
```
jrmoulton: Can you speak to the accuracy of doing this. Is this always true?

jrmoulton:
Thanks for doing this! Just one question

src/window/handle.rs
Fabi019: During testing I noticed that when setting `render_ready` to always true the `paint_state` was in pending for some cycles and then got to initialized. Without, the `paint_state` was always stuck in pending. While looking at it again now, I suspect the actual reason might be because the `last_presented_at` only gets set inside the condition. So when `renderer_ready` is false `render_frame` gets called without any/very low delay in between thus preventing the initialization. To confirm, I moved the `self.last_presented_at = Instant::now();` outside of the condition and with that it also works  with the original `renderer_ready` flag. Maybe that would be the better fix instead?

umbrageodotus: I think it would be but

Issue #505, opened 2024-07-08
Need a way to create a Floem NSView in a foreign Mac OS app


timboudreau:
Filing this as a placeholder, and will try to address it when I bubble back up out of DSP-code back to UI-land.

I'm attempting to use Floem as the UI for AudioUnit plugins for Mac OS using Apple's AudioUnit API.

These are libraries which are (optionally, but it's important for performance) loaded in-process by a host application (e.g. GarageBand, Logic, Adobe Audition).

They do not have ownership of the event loop or application (panic or exit would be *extremely* bad behavior).

The UI API is simply to provide a call that instantiates an NSView  - you implement whatever communication with the AudioUnit is needed under the hood).

So I need a way to instantiate an NSView which will be rendered and have its event handling done by Floem, without any assumptions about that *being* the entire application - and to be able to **unload** all of that cleanly, leaking nothing.

In the long run, implementing this would substantially increase the use cases for Floem on Mac OS, since it would be possible to write Floem *UI components* for use in **any** Mac OS application (possibly iOS too?) - and for most tasks, the performance is considerably better and the footprint much smaller than the equivalent UI using SwiftUI.

jrmoulton:
This should be pretty doable. I think we would need to switch to using the ui-events crate for events. then we would need to create a new delegate that would drive events that isn't a main app delegate and then just return a CAMetalLayer that we render into as an NSView.

timboudreau:
Coming back to this in earnest.  It seems like `winit` really wants to *own* the window and really, the main thread, period - and it seems from [this discussion](https://github.com/rust-windowing/winit/issues/1161) that the maintainers are not terribly interested in supporting attaching to a preexisting window.

But that is the use-case for audio plugins - your code is a library, and to open a UI, you get called with a native window handle to attach your event loop to and render into - there's no way around it.

There **is**, however, an alternative to winit designed for exactly this use-case: [baseview](https://github.com/RustAudio/baseview) - which has integrations for iced, egui and vizia - [nih-plug](https://github.com/robbert-vdh/nih-plug) contains working examples of usage.

So I am attempting to rip winit out from under Floem, based on a feature flag, and replace it with baseview.   It will be a *lot* of work, but it ought to be doable.

[...]

PR #1080, opened 2026-06-21
feat: add WindowIdExt::with_window_handle for native window handle access

zendy00:
## What

Adds `WindowIdExt::with_window_handle`, a backend-neutral way to obtain a window's native handle from a `WindowId`:

```rust
fn with_window_handle<F, T>(&self, f: F) -> Option<T>
where
    F: FnOnce(raw_window_handle::WindowHandle<'_>) -> T;
```

It runs `f` with the window's `raw_window_handle::WindowHandle`, returning `None` if there is no live OS window for the id (or the platform can't produce a handle). It wraps the existing internal `with_window` and mirrors the other `WindowIdExt` accessors (`scale`, `is_visible`, …), so it adds no new public types.

## Why

Embedders sometimes need the parent window's native handle to attach a native child — for example embedding a `wry` child webview via `WebViewBuilder::build_as_child(&handle)`. Today there is no public way to get it: the window registry (`with_window`) lives in the `pub(crate)` `window::tracking` module.

`raw_window_handle::WindowHandle` implements `HasWindowHandle`, so it plugs straight into `wry`/`build_as_child` and similar APIs. Exposing the `raw-window-handle` type (already a dependency) rather than the underlying `winit`/backend window keeps Floem from leaking its windowing-backend types — matching the design goal raised in #505 ("it would be a good thing if Floem doesn't directly expose winit's types"), and it should survive the planned winit → baseview abstraction since both back `HasWindowHandle`.

## Example

```rust
use floem::window::WindowIdExt;

window_id.with_window_handle(|handle| {
    wry::WebViewBuilder::new()
        .with_url("https://example.com")
        .build_as_child(&handle)
});
```

## Notes

- One method on the already-public `WindowIdExt`; no new public types, no `winit` leak.
- Compiles and `cargo clippy --lib` clean on current `main`.
- Used downstream to embed native browser tabs (a `wry` child webview) inside a Floem window.
- Related to the native-window interop discussed in #505 — this addresses the opposite direction (embedding a native child *into* a Floem window).


jrmoulton:
I think this looks reasonable. Thanks!

PR #1085, opened 2026-08-23
feat(window): expose WindowConfig::app_id() and consume launcher activation tokens

iam2r:
## Problem

Applications built with Floem currently never identify their windows to the
desktop environment:

1. **No application name is requested.** `create_window` builds winit's
   `WindowAttributes` without ever setting a name, so winit never calls
   `set_app_id()` on Wayland. On X11 the backend silently falls back to
   deriving `WM_CLASS` from `argv[0]`, masking the issue.

   Result on GNOME Wayland: the window cannot be associated with any
   `.desktop` entry — taskbars/docks show a generic icon or an "UNKNOWN"
   label, windows don't group, and launch tracking degrades
   (see lapce/lapce#2199).

2. **Launcher activation tokens are never consumed.** When a desktop shell
   launches an app it passes `XDG_ACTIVATION_TOKEN` (Wayland) /
   `DESKTOP_STARTUP_ID` (X11) and keeps the launch in a pending state until
   the token is used. Unconsumed, shells fall back to a timeout (~15s on
   GNOME), during which the dock shows a launching animation instead of the
   app entry.

Protocol capture of an affected app confirms no `set_app_id` is ever sent:

```
> WAYLAND_DEBUG=1 lapce |& grep app_id      # before: no output at all
```

## Change

This PR adds two small, opt-in capabilities to window creation:

- **`WindowConfig::app_id(impl Into<String>)`** — declares the application
  identity. When set, floem applies it via winit's
  `WindowAttributesExtWayland::with_name()` / `WindowAttributesExtX11::with_name()`
  (Wayland `app_id`, X11 `WM_CLASS`). When unset, behaviour is exactly as
  before.
- **Activation token consumption** — if the process was spawned by a desktop
  launcher, the token reported by
  `EventLoopExtStartupNotify::read_token_from_env()` is attached with
  `WindowAttributesExtStartupNotify::with_activation_token()`, letting the
  compositor complete the startup sequence immediately.

Both are `#[cfg(target_os = "linux")]`; other platforms are untouched.

## Why `lapce-backports`

Lapce pins this branch, so this is where the fix can reach the editor's next
release. I'm happy to port it to `main` as well once this lands or once the
direction is confirmed (main's new window-creation layout needs a slightly
different patch shape due to the `winit-core` split).

## Testing

- Built and run against Lapce on Zorin OS (GNOME, Wayland, 2x scale):
  - Before: dock/taskbar shows "UNKNOWN" / no icon; entry appears only after
    ~15s.
  - After: correct application icon immediately on window map.
- Protocol verification (`WAYLAND_DEBUG=1`) now shows
  `xdg_toplevel.set_app_id("dev.lapce.lapce")` before the surface maps.
- No unit tests added: the behaviour is platform integration (compositor +
  shell side). Happy to add builder-level tests if preferred.

### DIFF
--- a/src/app_handle.rs
+++ b/src/app_handle.rs
@@ -321,6 +321,7 @@ impl ApplicationHandle {
             mac_os_config,
             web_config,
             font_embolden,
+            app_id,
         }: WindowConfig,
     ) {
         let logical_size = size.map(|size| LogicalSize::new(size.width, size.height));
@@ -337,6 +338,23 @@ impl ApplicationHandle {
             .with_window_icon(window_icon)
             .with_resizable(resizable)
             .with_enabled_buttons(enabled_buttons);
+
+        // Set the application name so desktop environments can associate the
+        // window with its .desktop entry (dock icon, window grouping, etc).
+        // Without this, winit never calls set_app_id() on Wayland and windows
+        // have no identity at all.
+        #[cfg(target_os = "linux")]
+        if let Some(app_id) = app_id.as_deref() {
+            use winit::platform::x11::WindowAttributesExtX11;
+            window_attributes = WindowAttributesExtX11::with_name(window_attributes, app_id, app_id);
+        }
+        #[cfg(target_os = "linux")]
+        if let Some(app_id) = app_id.as_deref() {
+            use winit::platform::wayland::WindowAttributesExtWayland;
+            window_attributes =
+                WindowAttributesExtWayland::with_name(window_attributes, app_id, app_id);
+        }
+        }
 
         #[cfg(target_arch = "wasm32")]
         {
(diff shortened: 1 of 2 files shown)


adrianors:
Confirming this from a second downstream app: a floem 0.2.0 editor on GNOME Wayland (Arch Linux).

With no app_id, the running window shows up as a separate dock entry with a generic icon instead of grouping under its pinned .desktop launcher.

Patching floem-winit locally to send `xdg_toplevel.set_app_id("<bin name>")` (verified with `WAYLAND_DEBUG=1`) fixes the grouping and the icon immediately.

An opt-in `WindowConfig::app_id()` like this one would let us drop an XWayland workaround, which isn't viable on GL-only GPUs (wgpu's GL backend can fail to create a surface for an XWayland window). +1 for the `main` port too; happy to help test it.

PR #1087, opened 2026-09-18
Ignore listener registrations for removed views, and prune dead ids before layout broadcasts

criccomini:
## Problem

A view can be removed by an earlier update message in the same batch as its own `RegisterListener` message. The `EditorView` hits this easily because it calls `id.register_listener(LayoutChanged::listener_key())` when it is built, and an editor inside nested `dyn_container`s can be swapped out before that message is processed.

When that happens, the `RegisterListener` handler still pushes the dead `ViewId` into `window_state.listeners`, and records the key on the shared stale `ViewState` rather than the real one. `remove_view` has already run, so nothing ever cleans the id out. On the next `commit_box_tree`, `get_layout` on the dead id returns `Some` (the stale state owns a live taffy node), and `get_layout_window_origin` then calls `ViewStorage::box_tree`, which panics:

```
thread 'main' panicked at src/view/storage.rs:134:14:
all view ids are created with a root
   3: <floem::view::storage::ViewStorage>::box_tree
   4: ... <floem::view::id::ViewId>::get_layout_window_origin ...
   6: <floem::window::handle::WindowHandle>::commit_box_tree
```

## Fix

- `RegisterListener` skips ids whose view no longer exists.
- `commit_box_tree` prunes invalid ids out of the listener lists before the `LayoutChanged` and `VisualChanged` broadcasts.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


criccomini:
(This was Fable low, but the crash and debugging did happen :)

PR #1089, opened 2026-09-21
fix: forward post_paint through Box<dyn View>

criccomini:
`impl View for Box<dyn View>` forwards `post_paint` to the inner view's `paint`:

```rust
fn post_paint(&mut self, cx: &mut PaintCx) {
    (**self).paint(cx)
}
```

So a view wrapped in two boxes paints twice per frame, once in the paint pass and once in the post pass, and its own `post_paint` never runs.

Two boxes is the common case rather than the odd one. `into_any()` on an `AnyView` boxes it again, and `Container`, `Stack` and `dyn_stack` all call `into_any()` on their children, so any child handed over as an `AnyView` arrives double boxed. The paint traversal calls `post_paint` on `dyn View` without the trait in scope, so a single box dispatches straight to the concrete view and hides the bug; the second box dispatches to the `Box<dyn View>` impl and hits it. The inspector's overlay is drawn in `post_paint`, and any custom view that draws over its children after they paint depends on the hook running.

This forwards `post_paint` to `post_paint`. The new test paints a probe view behind one, two and three explicit boxes through the headless harness and counts each hook: once each with the fix, where before it was two paints and no post paint.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jrmoulton (approved):
Thanks!

criccomini:
@jrmoulton No problem! I actually have forked Floem because Opus/Fable have found so many bugs. I didn't want to inundate you guys with a ton of PRs. But if you're interested in it, my fork is in github.com/criccomini/floem under the `perch` branch. If you ask Fable/Opus 5.5 or Astra/Sol GPT 6 to go through the commits and compare with main, you're welcome to pull in whatever you think is relevant.

criccomini:
If you prefer, I can submit them as PRs indidivudally as well. I just didn't want to dump them all on you uninvited...

jrmoulton:
> I didn't want to inundate you guys with a ton of PRs

I'm short on time so I probably wouldn't be able to go through them very well anyways. but a list of bugs in a single issue would probably be a good thing if you / agent can do that. 


>  actually have forked Floem

I'm in a similar situation for my project. 

We're low on time but I dont think anyone involved here is opposed to AI usage so contributions (or lack) in whatever form are totally fine.

What are you using floem for?

